PR #123, opened 2015-12-28
feat(propertiesPanel): Add support for properties with tabular fields

coolngr8:
1. The Input Parameter, Output Parameter and Form Field contain fields which are tabular in nature(Map, List, Validation Constraints and Properties, for example)
2. Changed the PropertiesPanel.js to allow for such fields
3. Added InputParameterProps.js, OutputParameterProps.js and FormFieldProps.js under lib\provider\camunda\parts - these add additional property boxes to the property panel
4. Added Constraint.js, Entry.js, InputOutputParameter.js and Property.js - lib\provider\camunda\parts\implementation to support the newly added property boxes
5. Also, since only certain bpmn elements need these property boxes, the Camunda moddle file - camunda.json - needs to be modified to add abstract types that point to the appropriate bpmn elements. Have sent a separate pull request to accommodate same at camunda-bpmn-moddle

### DIFF
--- a/lib/PropertiesPanel.js
+++ b/lib/PropertiesPanel.js
@@ -82,6 +82,8 @@ function getFormControlValues(entryNode) {
   var values;
 
   var listContainer = domQuery('[data-list-entry-container]', entryNode);
+  var tableListContainer = domQuery('[data-list-table-container]', entryNode);
+
   if(!!listContainer) {
     values = [];
     var listNodes = domQuery.all('[data-list-entry-container] > div', entryNode);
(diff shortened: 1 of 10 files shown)


nikku:
Sorry for taking quite some time to review your contribution. Impressive stuff you have implemented.

However, looking into your changes we decided that we will not merge this huge PR for the following reasons:

* missing test cases
* missing support for redo / undo

Currently we are working on implementing all of that with the above things fixed.

Your PR will be an inspiration for us. Thanks for your efforts.

Issue #118, opened 2015-12-17
I can edit Generic Extensions


meyerdan:
Especially the `camunda:properties` that can be edited using the eclipse based modeler.

PR #142, opened 2016-02-12
pull request for camunda:Properties in bpmn extensions

evrenkutar:
hello,

Several days ago i created a title in bpmnio forum; https://forum.bpmn.io/t/extensionelements-camunda-properties-support/453 and i wrote a script to solve this problem. I was confused to name the file because there is so much usage for words [properties, extensions etc.]. I wrote code appropriate to your code styling and variable naming.

thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules/
 docs/api/
+.idea/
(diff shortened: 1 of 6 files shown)


romansmirnov:
Hi @evrenkutar,

Thanks for your pull request. We do not merge your pull request, since we we want to implement it differently (see https://github.com/bpmn-io/bpmn-js-properties-panel/issues/118).

Anyway, thanks for your effort.

Cheers,
Roman

PR #181, opened 2016-04-25
fix(providers): add explicit module dependencies to allow minification

### LINKED ISSUE #178: Providers break on minification
`BpmnPropertiesProvider` and `CamundaPropertiesProvider` are missing the `$inject` property, hence its not possible resolve their dependencies after minification (because params will be named like `a` or `b`).

I would expect something like

`Provider.$inject = ['eventBus', 'bpmnFactory', 'elementRegistry'];`

adrobisch:
fixes #178


nikku:
Merged via https://github.com/bpmn-io/bpmn-js-properties-panel/commit/1ead77ff4f9668df107d2f5e40c01ee1b8fcc65f. Thanks Andreas!

PR #203, opened 2016-08-19
feat(propeties-panel) add select box for business key to generated forms

aakhmerov:
related to #204

### DIFF
--- a/lib/provider/camunda/parts/FormProps.js
+++ b/lib/provider/camunda/parts/FormProps.js
@@ -10,7 +10,8 @@ var getBusinessObject = require('bpmn-js/lib/util/ModelUtil').getBusinessObject,
     formHelper = require('../../../helper/FormHelper'),
     utils = require('../../../Utils'),
     is = require('bpmn-js/lib/util/ModelUtil').is,
-    find = require('lodash/collection/find');
+    find = require('lodash/collection/find'),
+    each = require('lodash/collection/forEach');
 
 function generateValueId() {
   return utils.nextId('Value_');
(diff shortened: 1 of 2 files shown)


romansmirnov:
Review Hints:
* Why are the changes in `SelectEntryFactory` necessary?
* When I delete a form field which is declared as a business key field, then the attribute `camunda:businessKey` is still set to the id of the deleted form field. I would expect that the value is not set anymore in the model.
* The select box should not be visible, if there are no form field available. It feels a little bit strange, when there is an empty select box. It should be available if at least one form field available.
* The select box should only be visible, when a start event is selected on the diagram (and not in case of user task).
* If the type of a form field is set to `boolean`, then the form field should not be available as an option.
* Missing test cases: If I set/change the value of the business key, then the value should be present in the model. If I set the value of the business to `undefined`, then the value should not be set in the model anymore.

romansmirnov:
Review hints:
* use lodash filter instead of a for loop
* avoid duplicate call of `getExtensionElements(bo, 'camunda:FormData')`
* use `is()` to check the type of an element instead of `element.type`
* simplify boolean statement in `disabled`
* to unset the `businessKey` on the form data element use `undefined`
* add assertion inside `delete form field > on the business object`
* use the common pattern of writing test cases in the properties panel (ie. `on the business object` vs `in the DOM` and `execute`, `undo` and `redo`

romansmirnov:
Last remaining bug:
1) create two form fields `f1` and `f2`
2) select inside the business key select box `f1`
3) change type of `f2` to `boolean`
-> `f1` is not set as business key anymore

PR #208, opened 2016-08-30
Enable setting custom css classes for entry elements

albertodlh:
* As an Admin I can see a list of attributes.
* As an Admin I can create an attribute.


albertodlh:
Sorry. This was an error.

PR #226, opened 2017-02-18
fix(change): Stronger check for jQuery before unwraping.

juanmi108:
Other libraries exits that adds the `.get()` functionality to element nodes, in my case Mootools which behaves in a different fashion resulting in multiple errors.
 
Since the comment indicates that the condition is only for JQuery users, it should only fire if the lib is present.

### DIFF
--- a/lib/PropertiesPanel.js
+++ b/lib/PropertiesPanel.js
@@ -351,7 +351,7 @@ PropertiesPanel.prototype.attachTo = function(parentNode) {
   this.detach();
 
   // unwrap jQuery if provided
-  if (parentNode.get) {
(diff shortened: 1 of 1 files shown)


juanmi108:
Updated the check as suggested in https://github.com/bpmn-io/bpmn-js/pull/664

lib/PropertiesPanel.js
```diff
@@ -1098,7 +1098,7 @@ PropertiesPanel.prototype._createPanel = function(element, tabs) {
         }
 
         // unwrap jquery
-        if (html.get) {
+        if (html.get && window.jQuery) {
```
nikku: Still using the old check here?

PR #228, opened 2017-03-30
style(properties): resizable select fields

aldum:


nikku:
Could you provide a little bit more context, what you'd like to achieve with your PR and why?

aldum:
We sometimes need a lot of form fields in our BPMN models, and the default height is like 50 pixels, which makes it really frustrating to work with. This small modification makes select fields resizable, illustrated here:
http://i.imgur.com/5BTOdhs.png

nikku:
Merged via 87b9ef17abcbeb70b2c8343872a1216ff02afc67, thanks.

PR #229, opened 2017-04-03
Add scope parameter to allow creation of connector template

GuillaumeSmaha:


GuillaumeSmaha:
@nikku I create the PR to discuss about it.
It is not ready to merge.

GuillaumeSmaha:
@nikku I added the test for the functionnality.
Currently, there are  2 errors on `camunda::InputOutput` with `mail-task.json` test.

I don't know why these errors occur.
https://github.com/GuillaumeSmaha/bpmn-js-properties-panel/blob/ad2936b3b408d3d3a5027c45b7b72fefc2aabce9/test/spec/provider/camunda/element-templates/cmd/ChangeElementTemplateHandlerSpec.js#L193
https://github.com/GuillaumeSmaha/bpmn-js-properties-panel/blob/ad2936b3b408d3d3a5027c45b7b72fefc2aabce9/test/spec/provider/camunda/element-templates/cmd/ChangeElementTemplateHandlerSpec.js#L733
https://travis-ci.org/bpmn-io/bpmn-js-properties-panel/builds/218465265 

If I look at `mail-task.json`, I see 2 `camunda:inputParameter` and one `camunda:outputParameter`.
But in the test, there are 3 elements in the array.
Is it an error on master ?

GuillaumeSmaha:
Error fixed, I made a mistake in `mail-task.json`

StephenOTT:
@nikku any comments on this?

nikku:
Sorry for the long waiting time. Will look into this PR tomorrow.

nikku:
Updated your feature branch with changes from master.

Feedback incoming tomorrow.

GuillaumeSmaha:
@nikku I updated the branch from the master

lib/provider/camunda/element-templates/Validator.js
```diff
@@ -110,6 +112,32 @@ function Validator() {
       }
     }
 
+    if (scopes) {
+      if (!isArray(scopes)) {
+        err = this._logError('invalid scopes, should be scopes=[]', template);
+      }
+      else {
+        var _this = this;
```
nikku: Please declare at function root and call it `self` to comply with rest of code base.

lib/provider/camunda/element-templates/cmd/ChangeElementTemplateHandler.js
```diff
@@ -48,100 +48,174 @@ function ChangeElementTemplateHandler(modeling, commandStack, bpmnFactory) {
...
+          element: element,
+          businessObject: context,
+          properties: { inout: newInOut }
+        });
+      }
     }
   }
 
-  function updateExecutionListener(element, newTemplate) {
+  function updateExecutionListener(element, newTemplate, context, inside) {
```
nikku: Shouldn't it, based on whether context is given or not, be obvious whether we are inside?

test/spec/provider/camunda/element-templates/cmd/connector-task.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "ConnectorGetTask",
+  "id": "my.connector.http.get.Task",
+  "appliesTo": [
+    "bpmn:Task"
+  ],
+  "properties": [],
+  "scopes": [
+    {
```
nikku: When being used, this warns `no name given`.

nikku:
Any plans to act on the review feedback?

StephenOTT:
Yes.  Due to the delay in-between initial submission and feedback we had to move on to other work and devs are committed to another sprint at the moment.

nikku:
Thanks for the info and sorry again for the delay in reviewing.

GuillaumeSmaha:
@nikku

- I chose to use `scopes= []` because I didn't have a global vision of the project and I didn"t know if the document can contain multiple times the same key.
I will use `scope = {}` instead.

- For multiple `connector` scope, it will throw an error.

- The allowed scope are automaticly checked when the insertion command is executed. An error is  throwed when the document become invalid.

- I think `inside` can be removed from the users point of view and always set as true.

- I will add more test for setting scoped properties

GuillaumeSmaha:
@Nikku Edit for my previous note ; I remember why I used `inside`.

An element `<bpmn:extensionElements>` contains an element`<camunda:inputOutput>` and it was stored as a child in the code whereas a `<camunda:connector>` element contains an element`<camunda:inputOutput>` as a `properties` and not as a child. 

I think it will be better to change that. So it will be possible to delete `inside` var and also `insideConnector` var which used in https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/lib/helper/InputOutputHelper.js#L32
What do you think to remove `insideConnector` and to not do a special case for connector element ?

GuillaumeSmaha:
@nikku I updated the PR to use an object for `scopes` and I add test as you ask.

GuillaumeSmaha:
I tried to remove insideConnector but it needs a lot of work due to the difference between root element containing an extensionElement and camunda:Connector which contains inputOutput in properties

nikku:
Looking at the existing code base, we already have switches for similar behavior in place, cf. [field injection props](https://github.com/bpmn-io/bpmn-js-properties-panel/commit/b7a44cadaffc8efa66cc1f088679c4da9cba26af#diff-c0edb9115b9036c3b617a3277f769d2fL54) or the utility you've pointed out. Because of this I would be fine to just rename it to `insideConnector` for clarity.

Still, can't we assume `insideConnector` to be `true` when `context != null` is being passed ([ref](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/229#discussion-diff-113654973R153))?

GuillaumeSmaha:
> Still, can't we assume insideConnector to be true when context != null is being passed (ref)?

You have a better knowledge than me. I only focus on camunda:Connector,  do you know if there are others elements than Connector inside extensionElements ? If you think no, then I will made the change.

nikku:
I've pushed https://github.com/bpmn-io/bpmn-js-properties-panel/pull/229/commits/4a453815e30feb9dc7f5298afaa3654a8d420353 to your feature branch which fully removes insideConnector and switches to using _context_ only. I was able to remove [some of your hacks, too](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/229/commits/4a453815e30feb9dc7f5298afaa3654a8d420353#diff-d421d094011c806640c014c07ea9cdcbL176).

Please verify that stuff is still working from your side.

nikku:
(You can simply retrieve my changes by pulling your feature branch).

lib/provider/camunda/element-templates/parts/CustomProps.js
```diff
@@ -139,7 +139,7 @@ module.exports = function(element, elementTemplates, bpmnFactory, translate) {
 
         id = 'custom-' + template.id + '-' + scopeIdx + '-' + idx;
 
-        entry = renderCustomField(id, p, idx, scope);
+        entry = renderCustomField(id, p, idx, scopeIdx);
```
nikku: TODO: Shouldn't we have a test case for this?

GuillaumeSmaha:
I tried with this examples and it works ! 👍 

```[json]
    "scopes": {
      "camunda:Connector": {
        "insideProperties": true,
        "properties": [
          {
            "label": "ConnectorId",
            "type": "String",
            "value": "My Connector HTTP",
            "binding": {
              "type": "property",
              "name": "connectorId"
            },
            "constraints": {
              "notEmpty": true
            }
          },
          {
            "label": "URL",
            "type": "String",
            "binding": {
              "type": "camunda:inputParameter",
              "name": "recipient"
            },
            "constraints": {
              "notEmpty": true
            }
          },
          {
            "label": "Method",
            "type": "Dropdown",
            "choices": [
              {
                "value": "GET",
                "name":"GET"
              },
              {
                "value": "POST",
                "name": "POST"
              },
              {
                "value": "PUT",
                "name": "PUT"
              },
              {
                "value": "PATCH",
                "name": "PATCH"
              },
              {
                "value": "DELETE",
                "name": "DELETE"
              }
            ],
            "binding": {
              "type": "camunda:inputParameter",
              "name": "method"
            },
            "constraints": {
              "notEmpty": true
            }
          },
          {
            "type": "Hidden",
            "value": "Camunda",
            "binding": {
              "type": "camunda:inputParameter",
              "name": "agent"
            }
          },
          {
            "label": "Response",
            "type": "String",
            "value": "${S(response)}",
            "binding": {
              "type": "camunda:outputParameter",
              "name": "wsResponse",
              "source": "wsResponse"
            }
          }
        ]
      }
    }
  },
````

**Comment by nikku** [comment]
We're getting closer to merging...

**Comment by nikku** [comment]
While working on a few more improvements I recognized that the example connector `CustomProps.bpmn` you've added is invalid: It does not contain a value for a property that, by template definition, must be provided.

As a result one of the test cases fails unless we apply [another fix](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/229/commits/e18e33d412fe9deae84be641ff0b8fe7ed9c1cb1).

My question is: Should we deal with broken (without value) connector definitions? If yes, we'd need to pull in https://github.com/bpmn-io/bpmn-js-properties-panel/pull/229/commits/e18e33d412fe9deae84be641ff0b8fe7ed9c1cb1. Otherwise you should change the `CustomProps.bpmn` and provide actual values.

**Comment by nikku** [comment]
One last thing I'd like you to look into:

As a pattern in `CustomProps`, we'll always create the whole model tree when setting an individual property. That said, when the user provides an incomplete BPMN 2.0 XML reading and setting properties should still work for individual entries. An example XML could be this:

```xml
<bpmn:ServiceTask id="ConnectorTask" name="Connector Task" camunda:modelerTemplate="my.connector.Task">
</bpmn:ServiceTask>
```

When the user changes a property such as `url`, `CustomProps` does make sure the respective model structure is created: 

```xml
<bpmn:ServiceTask id="ConnectorTask" name="Connector Task" camunda:modelerTemplate="my.connector.Task">
  <bpmn:extensionElements>
    <camunda:connector>
      <camunda:inputOutput>
        <camunda:inputParameter name="url"></camunda:inputParameter>
      </camunda:inputOutput>
      <camunda:connectorId>My Connector HTTP</camunda:connectorId>
    </camunda:connector>
  </bpmn:extensionElements>
</bpmn:ServiceTask>
```


Your implementation right now assumes that at least the following stub is always given:

```xml
<bpmn:ServiceTask id="ConnectorTask_NoData" name="Connector Task NoData" camunda:modelerTemplate="my.connector.Task">
  <bpmn:extensionElements>
    <camunda:connector>
      <camunda:inputOutput />
    </camunda:connector>
  </bpmn:extensionElements>
</bpmn:ServiceTask>
```

Could you update this to create the whole structure on the fly, in case it does not exist yet?

felix-mueller:
Hi @GuillaumeSmaha
any chance to check Nikku's proposal? I was also looking for element templates for connectors today so maybe we can move this forward?

GuillaumeSmaha:
@nikku @felix-mueller I will work on it in the next days

firle:
Hey Guys!
Currently we developing with Camunda and to simplify our workflows it would be great to use templates for connectors. Since the last reply is 3 weeks ago i want to ask how long it will take to include this feature discussed here to include on the master branch. This will push us forward!
Thanks and Thanks to your work guys!

GuillaumeSmaha:
I am still working on this on my free time.

GuillaumeSmaha:
@nikku I fixed CustomProp to generate the whole model tree

nikku:
Thanks for the changes. I'll have a look.

nikku:
Yep. CI is failing for this PR. I'll have a look, as it does not seem to be related to the added functionality.

GuillaumeSmaha:
@nikku @StephenOTT I fix CI test (Merge and Fix for syntax)

nikku:
I've consolidated your changes on the [`add-scopes-snapshot` branch](https://github.com/bpmn-io/bpmn-js-properties-panel/commits/add-scopes-snapshot). There are a couple of issues with our CI, unrelated to this PR. I'm about to fix these and we're ready to merge. 

In general I'd ask you to follow our [Contribution Guidelines](https://github.com/bpmn-io/bpmn-js/blob/master/CONTRIBUTING.md#contributing-improvements) to make reviewing and merging easier. This includes running `npm run all` to ensure code is properly formated and working against our test suite, providing commit messages adhering to our conventions and using the rebase workflow instead of merge commits to pull in upstream changes (read more about it [here](https://www.gun.io/blog/how-to-github-fork-branch-and-pull-request)).

nikku:
Closes https://github.com/bpmn-io/bpmn-js-properties-panel/pull/229.

felix-mueller:
Awesome, thank you @GuillaumeSmaha and @nikku !

GuillaumeSmaha:
@nikku Thanks :)

StephenOTT:
@GuillaumeSmaha Great work! 🎉 

@nikku would the new feature be in the nightly of Camunda Modeler?

PR #231, opened 2017-04-04
feat(element-templates): Add Map and List type for inputParameter and outputParameter

GuillaumeSmaha:
Issue: https://github.com/camunda/camunda-modeler/issues/497

- [x] Users are able to configure Maps as a type for input/output mappings
- [ ] The modeler displays an editor for key/values configured as part of the map similar to the map editor in 

Not ready to merge.
There is missing the list and map editor in the panel

### DIFF
--- a/lib/provider/camunda/element-templates/CreateHelper.js
+++ b/lib/provider/camunda/element-templates/CreateHelper.js
@@ -7,21 +7,55 @@
  * @param {PropertyBinding} binding
  * @param {String} value
  * @param {BpmnFactory} bpmnFactory
+ * @param {String} type
  *
  * @return {ModdleElement}
  */
-function createInputParameter(binding, value, bpmnFactory) {
+function createInputParameter(binding, value, bpmnFactory, type) {
   var scriptFormat = binding.scriptFormat,
       parameterValue,
       parameterDefinition;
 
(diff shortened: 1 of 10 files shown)


GuillaumeSmaha:
@nikku Do you have any advices to implement the GUI part ?

nikku:
The GUI may become extremely complex quite easily, depending on what you'd like to support. Are bindings dynamic necessary and if so, in which sense? Would you like to support:

* single level of map / list?
* nested?

For the map, should users be able to 

* provide loose collection of key/value pairs
* values only
* are values plain strings, or can they contain nested complex objects?

For the list, should users be able to specify list entries in individual inputs _or_ as coma separated lists?

nikku:
Do you still plan on completing this PR?

GuillaumeSmaha:
@nikku Sorry, for my answer delay, I was in vacation. Currently, I haven't a lot of time since I changed company.

nikku:
Thanks for the response and best of luck in the new company.

adrianriepl:
Is it possible to complete this PR so that the first part of the feature can be used? 

This way element templates can be created for the `http-connector`. The displayed editor in the modeler could be implemented in the near future.

PR #239, opened 2017-09-04
fix(select-entry-factory): Add prefix to select box css classes

### LINKED ISSUE #238: SelectEntryFactory: Missing css class prefix
The select entry factory sets the css class to "dropdown" without the "bpp-" prefix.
This clashes with some css frameworks.

ColinFrick:
Fixes #238


nikku:
Merged via https://github.com/bpmn-io/bpmn-js-properties-panel/commit/122e2f1ea3399bdb094bf510dbdc208dd7332d9a.

PR #241, opened 2017-09-07
fix(select-entry-factory): Add prefix to select css class

### LINKED ISSUE #240: TextBox: Empty fields in Firefox
In Firefox empty textboxes are only 6 pixels high (equal to the padding):

image

Tested with 55.0.2 (64-bit) and 56.0b9 (64-bit)

ColinFrick:
Fixes #240


ColinFrick:
Sorry wrong commit message

PR #244, opened 2017-10-12
fix(facotry): TextInputEntryFactory handles disable property correctly

### LINKED ISSUE #245: options.disabled for TextInputEntryFactory does not work
### Expected Behavior

`TextInputEntryFactory` should create disabled fields when passing the option `disabled: function() { return true; }`.

#### First attempt

I experienced two issues which kind of seemed to be non-deterministic to me.
I encountered the issue first in a project of mine - which I won't link here because it's very big.
There, I added a custom properties panel field for `bpmn:name` which should be disabled by adding `disabled: function() { return true; }` to the options.
In my project following things happened:
1. Label and content of the field are colored in light-grey
2. The corresponding HTML element appears to be configured correctly and appears as:
```html
<div class="bpp-field-wrapper" data-disable="isDisabled" disabled="">
```
3. You can enter values, but they get reset after a short amount of time.
However: The name label of the respective element is set to the value you have entered.
The business object remains - correctly - with no attribute `bpmn:name` set.

#### Second attempt

I also tried to reproduce the error in a more simple environment building up on [the example for a properties panel extension](https://github.com/bpmn-io/bpmn-js-examples/tree/master/properties-panel-extension).
There I added a file `app/provider/magic/parts/AltNameProps.js` which replaced the built-in `NameProps` and has the following content:
```js
'use strict';

var entryFactory = require('bpmn-js-properties-panel/lib/factory/EntryFactory'),
    is = require('bpmn-js/lib/util/ModelUtil').is;

module.exports = function(group, element) {

  if (!is(element, 'bpmn:Collaboration')) {

    var options = {
        id: 'name',
        label: 'Name',
        modelProperty: 'name',
        disabled: function() { return true; }
    };
    if (is(element, 'bpmn:TextAnnotation')) {
      options.modelProperty = 'text';
    }

    group.entries.push(entryFactory.textField(options));

  }
};
```
I also tried this with `modelProperty: 'bpmn:name'`.

Now to the results. With `modelProperty` set to `'name'`:
1. The label of the field is colored in light-grey.
2. The content of the field is not colored in light-grey.
3. When entering a value it gets stored and the label is applied to the rendered element.
4. As in my personal project, the HTML seems good.

With `modelProperty` set to `'bpmn:name'`:
1. The label of the field is colored in light-grey.
2. The content of the field is not colored in light-grey.
3. When entering a value it gets stored.
4. The stored value is only applied to the respective rendered element if this is a Task.
5. As in my personal project, the HTML seems good.

All in all - this is strange behavior and disabling a text field doesn't seem to work.

dominikhrn:
Fixes #245 
-TextInputEntryFactory didn't use disable for input tag.
-Button used data-disable**d** -> changed to data-disable

### DIFF
--- a/lib/factory/TextInputEntryFactory.js
+++ b/lib/factory/TextInputEntryFactory.js
@@ -44,10 +44,10 @@ var textField = function(options, defaultParameters) {
       (canBeDisabled ? 'data-disable="isDisabled"' : '') +
       (canBeHidden ? 'data-show="isHidden"' : '') +
       '>' +
-      '<input id="camunda-' + resource.id + '" type="text" name="' + options.modelProperty+'" ' +
+      '<input id="camunda-' + resource.id + '" ' + (canBeDisabled ? 'data-disable="isDisabled"' : '') + ' type="text" name="' + options.modelProperty+'" ' +
         ' />' +
       '<button class="' + actionName + '" data-action="' + actionName + '" data-show="' + showName + '" ' +
-        (canBeDisabled ? 'data-disabled="isDisabled"' : '') +
+        (canBeDisabled ? 'data-disable="isDisabled"' : '') +
         (canBeHidden ? ' data-show="isHidden"' : '') + '>' +
         '<span>' + buttonLabel + '</span>' +
       '</button>' +


nikku:
Thanks for your contribution. It so happened that someone else filed a PR for this one, too. 

Closed via https://github.com/bpmn-io/bpmn-js-properties-panel/commit/f9bdda81ff90203f70ee53c3ecba492e5bbe0feb and https://github.com/bpmn-io/bpmn-js-properties-panel/commit/b3d7a778a5254c0ada8bba822cd67980335e4567.

PR #257, opened 2018-03-12
Export setPropertyValue from element templates helpers

adamscybot:
### Proposed Changes

I'm creating my own properties panel in React with element template support. It is backed by these core getters and setters. `getPropertyValue` is already exported whilst `setPropertyValue` is not. This just exports the setter so it can be used directly if needed.


nikku:
Thanks for the PR.

nikku:
Do you mind sharing your built-with-react properties panel? I know many users request such thing.

nikku:
In case you want to contribute back: Drop us a line in [our forum](https://forum.bpmn.io/) or file a PR against our [example repository](https://github.com/bpmn-io/bpmn-js-examples).

PR #272, opened 2018-07-31
refactor(properties-panel): cleaned up variables and used domAttr()

froissant:
Changes to harmonize the code

### Proposed Changes
* Cleaned up variables to match the "click" event handler
* Replaced the usage of getAttribute() by domAttr()


nikku:
Merged via d3d2c7c4682a15edff0984433d9fbd415ca45b36, thank you.

PR #276, opened 2018-09-12
fix(PropertiesPanel): activate first tab when active tab was removed

froissant:
Fixes an issue where the panel was activating a tab that was removed; in case the saved active tab doesn't exists, activate the first tab.

### DIFF
--- a/lib/PropertiesPanel.js
+++ b/lib/PropertiesPanel.js
@@ -424,6 +424,12 @@ PropertiesPanel.prototype.activateTab = function(tab) {
 
     domClasses(tabLinkNode).toggle('bpp-active', tabId === currentTabId);
   });
+
+  // verify that the activate tab was found, otherwise activate the first tab
+  var activeTabNode = domQuery('.bpp-properties-tab.bpp-active', panelNode);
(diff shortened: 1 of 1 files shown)


nikku:
Could you elaborate a bit on your use case?

Calling `PropertiesPanel#activateTab` with a non-existing tab is a programming error that can easily be avoided by the API caller.

froissant:
Hello,
Use case: you have different tabs on different elements (for examples, on an element of type *Task I have two tabs: "Common Properties" and "Task Properties"; on *Events I have two tabs "Common Properties" and "Event Properties").

If I am on an Event element, with the "Events Properties" tab active, and I click a Task element, then currently no tab will be shown as the "Events Properties" activeTabId will be saved and passed to the activateTab method (line 477) -- since it doesn't exists on the new element, nothing will be shown in the property panel. The fix is to display the first available tab in this case.

nikku:
Closing this issue, as we are still not able to understand this / reproduce this as a bug.

nikku:
We'd be happy to re-open it if you provide us with exact steps to reproduce the issue.

It seems like things work already as intended as shown in the following screen cast:

foo

PR #279, opened 2018-10-31
Show correct name properties

### LINKED ISSUE #278: Show correct name properties
In the [forum](https://forum.bpmn.io/t/name-in-group-element-not-rendered/112) it was discussed that group names not rendered correctly. https://github.com/bpmn-io/bpmn-js/pull/880 solved it. 

However the properties panel should not display the `name` property for this kind of element, since the [BPMN 2.0 specification](https://www.omg.org/spec/BPMN/2.0/PDF) (p. 68-69 "Group") introduces that groups can't have a name itself (just via the related category value, as mentioned in the forum). Furthermore it should be more consistent with the specification if there is a property `categoryValue` or whatever. 

Another example: The TextAnnotation element has also a property `name`, which is mapped to the specification-correct property `text` afterwards. It would be a lot more consistent with the bpmn specs if the correct property-name is displayed. 

### Expected Behavior
The `name` property for following element-types should be displayed correctly:

- `bpmn:Group`: `Name` ---> `Category Value`
- `bpmn:TextAnnotation`: `Name` ---> `Text`

### Actual Behavior
The property `name` is displayed, which is not correct according the BPMN 2.0. specification

image

### Steps to reproduce the Behavior

- Open a diagram with a group in the Camunda Modeler, e.g. [group_name.bpmn](https://github.com/bpmn-io/bpmn-js-properties-panel/files/2510356/group_name.bpmn.txt)
- Open Properties Panel

pinussilvestrus:
Fixes #278 

* Show correct name-labels for `bpmn:TextAnnotation` (Text) and `bpmn:Group` (Category Value)
* Make `bpmn:CategoryValue` for `bpmn:Group` editable in properties panel
* Initializes `bpmn:CategoryValue` if needed, via `CategoryHelper` (depends on  https://github.com/bpmn-io/bpmn-js/pull/1040)

### DIFF
--- a/lib/provider/bpmn/parts/NameProps.js
+++ b/lib/provider/bpmn/parts/NameProps.js
@@ -9,7 +9,7 @@ module.exports = function(group, element, translate) {
 
     var options;
     if (is(element, 'bpmn:TextAnnotation')) {
-      options = { modelProperty: 'text' };
+      options = { modelProperty: 'text', label: 'Text' };
     }
 
     // name
--- a/lib/provider/bpmn/parts/NameProps.js
+++ b/lib/provider/bpmn/parts/NameProps.js
@@ -1,15 +1,58 @@
 'use strict';
 
 var nameEntryFactory = require('./implementation/Name'),
-    is = require('bpmn-js/lib/util/ModelUtil').is;
+    createCategoryValue = require('../../../helper/CategoryHelper').createCategoryValue,
+    is = require('bpmn-js/lib/util/ModelUtil').is,
+    getBusinessObject = require('bpmn-js/lib/util/ModelUtil').getBusinessObject;
 
-module.exports = function(group, element, translate) {
+module.exports = function(group, element, bpmnFactory, canvas, translate) {
+
+  function initializeCategory(semantic) {
+    var rootElement = canvas.getRootElement(),
+        definitions = getBusinessObject(rootElement).$parent,
+        categoryValue = createCategoryValue(definitions, bpmnFactory);
+
+    semantic.categoryValueRef = categoryValue;
+
+  }
+
+  function setGroupName(element, values) {
+    var bo = getBusinessObject(element),
+        categoryValueRef = bo.categoryValueRef;
+
+    if (!categoryValueRef) {
+      initializeCategory(bo);
+    }
+
+    // needs direct call to update categoryValue properly
+    return {
+      cmd: 'element.updateLabel',
+      context: {
+        element: element,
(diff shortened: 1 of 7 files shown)


barmac:
When I change the category value, it is not rendered immediately. I need to change ID or documentation to trigger rerendering. See attached GIF file for reference.

kapture 2018-11-19 at 11 17 33

pinussilvestrus:
@barmac Thanks for reviewing!
It seems that `cmdHelper.updateBusinessObject` not triggers the rerendering properly. I'll check it later.

pinussilvestrus:
Updated with a fix which triggers the rerendering of the `bpmn:Group` shape

lib/provider/bpmn/parts/NameProps.js
```diff
@@ -1,15 +1,40 @@
...
-    is = require('bpmn-js/lib/util/ModelUtil').is;
+    is = require('bpmn-js/lib/util/ModelUtil').is,
+    getBusinessObject = require('bpmn-js/lib/util/ModelUtil').getBusinessObject,
+    cmdHelper = require('../../../helper/CmdHelper');
 
 module.exports = function(group, element, translate) {
 
+  function setGroupName(element, values) {
+    var bo = getBusinessObject(element),
+        categoryValueRef = bo.categoryValueRef || {};
```
nikku: This is not sufficient to instantiate a missing `categoryValueRef`.

We must 

* properly create the category via the `BpmnFactory`
* add the category in the respective place
* link it to the BO if it does not exist

test/spec/provider/bpmn/ElementNameSpec.js
```diff
@@ -370,6 +371,93 @@ describe('element-name-properties', function() {
 
     });
 
+    describe('of a group', function() {
```
nikku: TODO: 

* Verify that `no name` -> `name is set` -> `CategoryValue` is created and properly linked to the document model.

pinussilvestrus:
I think it's worth it to wait here for the overall group-modeling implementation (cf. https://github.com/bpmn-io/bpmn-js/issues/343). After we have a detailed understanding of how to model groups it makes sense to finish the properties panel implementation.

pinussilvestrus:
For category initialization, we can use the implementation introduced in https://github.com/bpmn-io/bpmn-js/pull/1017.

nikku:
@barmac Could you have a look into this one?

barmac:
> @barmac Could you have a look into this one?

Sure, I'll do this.

PR #285, opened 2018-11-23
Default attr task stuff

nikku:
Correctly handles `camunda:isStartableInTasklist` default parameter


nikku:
@philippfromme Could you have a look into this one?

PR #292, opened 2019-02-19
chore(PropertiesPanel): fix typo in documentation

StanleySweet:
### Proposed Changes

* Fix a typo in the documentation Booelan -> Boolean
* Add a final eol

### DIFF
--- a/lib/PropertiesPanel.js
+++ b/lib/PropertiesPanel.js
@@ -487,7 +487,7 @@ PropertiesPanel.prototype.update = function(element) {
  *
  * @param  {Object} current
  * @param  {Object} newTabs
- * @return {Booelan}
+ * @return {Boolean}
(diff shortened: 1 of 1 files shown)


pinussilvestrus:
Really thanks for your contribution! 🎉 As you can see inside our [commit history](https://github.com/bpmn-io/bpmn-js-properties-panel/commits/master), we use a message guideline to label our commits. Maybe you can adapt your commit to be in this form? It would be something like:

`chore(PropertiesPanel): fix typo in documentation`

You can have a further look in the commit message guideline via [this documentation](https://www.conventionalcommits.org/en/v1.0.0-beta.2/) which comes near to ours.

StanleySweet:
Mmmh looks like I have to create another one.

PR #295, opened 2019-03-14
Localize literals, which are not yet localized.

rumata28:
- use "translate" where it is available
- Script.js changed to accept "translate" as an input parameter of exported function
  - Changed 3 places, which used the "Script.js", so that the "translate" module passed as a parameter down to function as a parameter value
- Listener.js - refactored, so that translation would be obtained on rendering, not on module loading.  Language switching would work correctly that way.

Closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/297

### DIFF
--- a/lib/provider/bpmn/parts/EventProps.js
+++ b/lib/provider/bpmn/parts/EventProps.js
@@ -33,11 +33,11 @@ module.exports = function(group, element, bpmnFactory, elementRegistry, translat
           signalEventDefinition = eventDefinitionHelper.getSignalEventDefinition(element);
 
       if (messageEventDefinition) {
-        message(group, element, bpmnFactory, messageEventDefinition);
+        message(group, element, bpmnFactory, messageEventDefinition, translate);
       }
 
       if (signalEventDefinition) {
-        signal(group, element, bpmnFactory, signalEventDefinition);
+        signal(group, element, bpmnFactory, signalEventDefinition, translate);
       }
 
     }
@@ -45,7 +45,7 @@ module.exports = function(group, element, bpmnFactory, elementRegistry, translat
 
   // Special Case: Receive Task
   if (is(element, 'bpmn:ReceiveTask')) {
-    message(group, element, bpmnFactory, getBusinessObject(element));
+    message(group, element, bpmnFactory, getBusinessObject(element), translate);
   }
 
   // Error Event Definition
@@ -66,7 +66,8 @@ module.exports = function(group, element, bpmnFactory, elementRegistry, translat
         var showErrorCodeVariable = isCatchingErrorEvent,
             showErrorMessageVariable = isCatchingErrorEvent;
 
-        error(group, element, bpmnFactory, errorEventDefinition, showErrorCodeVariable, showErrorMessageVariable);
+        error(group, element, bpmnFactory, errorEventDefinition, showErrorCodeVariable, showErrorMessageVariable,
+          translate);
       }
     }
   });
@@ -88,7 +89,8 @@ module.exports = function(group, element, bpmnFactory, elementRegistry, translat
       var escalationEventDefinition = eventDefinitionHelper.getEscalationEventDefinition(element);
 
       if (escalationEventDefinition) {
-        escalation(group, element, bpmnFactory, escalationEventDefinition, showEscalationCodeVariable);
+        escalation(group, element, bpmnFactory, escalationEventDefinition, showEscalationCodeVariable,
+          translate);
       }
     }
 
@@ -108,7 +110,7 @@ module.exports = function(group, element, bpmnFactory, elementRegistry, translat
       var timerEventDefinition = eventDefinitionHelper.getTimerEventDefinition(element);
 
       if (timerEventDefinition) {
-        timer(group, element, bpmnFactory, timerEventDefinition);
+        timer(group, element, bpmnFactory, timerEventDefinition, translate);
       }
     }
   });
@@ -126,13 +128,13 @@ module.exports = function(group, element, bpmnFactory, elementRegistry, translat
       var compensateEventDefinition = eventDefinitionHelper.getCompensateEventDefinition(element);
 
       if (compensateEventDefinition) {
-        compensation(group, element, bpmnFactory, compensateEventDefinition, elementRegistry);
+        compensation(group, element, bpmnFactory, compensateEventDefinition, elementRegistry, translate);
       }
     }
   });
 
 
-  // Conditional Event Defintion
+  // Conditional Event Definition
   var conditionalEvents = [
     'bpmn:StartEvent',
     'bpmn:BoundaryEvent',
@@ -146,7 +148,7 @@ module.exports = function(group, element, bpmnFactory, elementRegistry, translat
     var conditionalEventDefinition = eventDefinitionHelper.getConditionalEventDefinition(element);
 
     if (conditionalEventDefinition) {
-      condition(group, element, bpmnFactory, conditionalEventDefinition, elementRegistry);
+      condition(group, element, bpmnFactory, conditionalEventDefinition, elementRegistry, translate);
     }
   }
 
--- a/lib/provider/bpmn/parts/implementation/ConditionalEventDefinition.js
+++ b/lib/provider/bpmn/parts/implementation/ConditionalEventDefinition.js
@@ -6,7 +6,7 @@ var entryFactory = require('../../../../factory/EntryFactory'),
 var is = require('bpmn-js/lib/util/ModelUtil').is,
     isEventSubProcess = require('bpmn-js/lib/util/DiUtil').isEventSubProcess;
 
-module.exports = function(group, element, bpmnFactory, conditionalEventDefinition) {
+module.exports = function(group, element, bpmnFactory, conditionalEventDefinition, elementRegistry, translate) {
 
   var getValue = function(modelProperty) {
     return function(element) {
@@ -29,8 +29,8 @@ module.exports = function(group, element, bpmnFactory, conditionalEventDefinitio
   };
 
   group.entries.push(entryFactory.textField({
-    id : 'variableName',
-    label : 'Variable Name',
+    id: 'variableName',
+    label: translate('Variable Name'),
     modelProperty : 'variableName',
 
     get: getValue('variableName'),
@@ -42,9 +42,9 @@ module.exports = function(group, element, bpmnFactory, conditionalEventDefinitio
 
   if (!isConditionalStartEvent) {
     group.entries.push(entryFactory.textField({
-      id : 'variableEvent',
-      label : 'Variable Event',
-      description: 'Specify more than one variable change event as a comma separated list.',
+      id: 'variableEvent',
+      label: translate('Variable Event'),
+      description: translate('Specify more than one variable change event as a comma separated list.'),
       modelProperty : 'variableEvent',
 
       get: getValue('variableEvent'),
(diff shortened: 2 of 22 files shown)


rumata28:
I also changed in some places `'Must provide a value.'` to `translate('Must provide a value') + '.'`.
The rationale of that is to have just one literal to localize: `Must provide a value` instead of 2 almost equal literals: `Must provide a value` and `Must provide a value.`.
In most of other places in the project you have `Must provide a value` - without dot at the end.

lib/provider/bpmn/parts/implementation/ConditionalEventDefinition.js
```diff
@@ -42,8 +42,8 @@ module.exports = function(group, element, bpmnFactory, conditionalEventDefinitio
 
   if (!isConditionalStartEvent) {
     group.entries.push(entryFactory.textField({
-      id : 'variableEvent',
-      label : 'Variable Event',
+      id: 'variableEvent',
+      label: translate('Variable Event'),
       description: 'Specify more than one variable change event as a comma separated list.',
```
barmac:
```suggestion
       description: translate('Specify more than one variable change event as a comma separated list.'),
```

lib/provider/bpmn/parts/EventProps.js
```diff
@@ -88,7 +88,7 @@ module.exports = function(group, element, bpmnFactory, elementRegistry, translat
       var escalationEventDefinition = eventDefinitionHelper.getEscalationEventDefinition(element);
 
       if (escalationEventDefinition) {
-        escalation(group, element, bpmnFactory, escalationEventDefinition, showEscalationCodeVariable);
+        escalation(group, element, bpmnFactory, escalationEventDefinition, showEscalationCodeVariable, translate);
```
barmac:
```suggestion
escalation(group, element, bpmnFactory, escalationEventDefinition, showEscalationCodeVariable,
  translate);
```

lib/provider/bpmn/parts/EventProps.js
```diff
@@ -66,7 +66,7 @@ module.exports = function(group, element, bpmnFactory, elementRegistry, translat
         var showErrorCodeVariable = isCatchingErrorEvent,
             showErrorMessageVariable = isCatchingErrorEvent;
 
-        error(group, element, bpmnFactory, errorEventDefinition, showErrorCodeVariable, showErrorMessageVariable);
+        error(group, element, bpmnFactory, errorEventDefinition, showErrorCodeVariable, showErrorMessageVariable, translate);
```
barmac:
```suggestion
        error(group, element, bpmnFactory, errorEventDefinition, showErrorCodeVariable, showErrorMessageVariable,
          translate);
```
Let's try to not exceed 100 characters per line.

lib/provider/camunda/parts/VariableMappingProps.js
```diff
@@ -313,13 +318,13 @@ module.exports = function(group, element, bpmnFactory, translate) {
...
           validation.source =
-            'Mapping must have a ' + (values.sourceLabel.toLowerCase() || 'value');
+            translate('Mapping must have a') + ' ' + (values.sourceLabel.toLowerCase() || 'value');
         }
 
         var inOutType = getInOutType(mapping);
 
         if (WHITESPACE_REGEX.test(values.source) && inOutType !== 'sourceExpression') {
-          validation.source = values.sourceLabel + ' must not contain whitespace';
+          validation.source = values.sourceLabel + ' ' + translate('must not contain whitespace');
```
barmac:
```suggestion
          validation.source = translate('{label} must not contain whitespace', { label: values.sourceLabel });
```
Please use template replacements, as the variable may not be allowed at first position in a localized version. See https://github.com/bpmn-io/diagram-js/blob/master/lib/i18n/translate/translate.js for reference.

lib/provider/camunda/parts/VariableMappingProps.js
```diff
@@ -313,13 +318,13 @@ module.exports = function(group, element, bpmnFactory, translate) {
       if (mapping) {
         if (!values.source) {
           validation.source =
-            'Mapping must have a ' + (values.sourceLabel.toLowerCase() || 'value');
+            translate('Mapping must have a') + ' ' + (values.sourceLabel.toLowerCase() || 'value');
```
barmac:
```suggestion
          validation.source = values.sourceLabel ?
            translate('Mapping must have a {value}', { value: values.sourceLabel.toLowerCase() }) :
            translate('Mapping must have a value');
```
Unfortunately, for now there is no easy way to add default values, but we should do it this way.

lib/provider/camunda/parts/VariableMappingProps.js
rumata28: Thanks!

lib/provider/camunda/parts/VariableMappingProps.js
barmac: It seems like this suggestion block does not apply for multilines 🙄.

lib/provider/camunda/parts/implementation/Callable.js
```diff
@@ -234,7 +234,7 @@ module.exports = function(element, bpmnFactory, options, translate) {
     validate: function(element, values, node) {
       var elementRef = values.callableElementRef;
       var type = getCallableType(element);
-      return isSupportedCallableType(type) && !elementRef ? { callableElementRef: 'Must provide a value.' } : {};
+      return isSupportedCallableType(type) && !elementRef ? { callableElementRef: translate('Must provide a value') + '.' } : {};
```
barmac:
```suggestion
      return isSupportedCallableType(type) && !elementRef ? { callableElementRef: translate('Must provide a value.') } : {};
```
Let's not exclude any parts of template strings. Some languages use own fullstop instead of common one.

lib/provider/camunda/parts/VariableMappingProps.js
rumata28: ok.  Just wanted to have it consistent over the component.
In many similar places, the message is just "Must provide a value", with no dot.

Can we just omit the dot?

barmac: We can think about it in a separate PR, as this is simply not a part of i18n, ok?

lib/provider/camunda/parts/implementation/Listener.js
```diff
@@ -55,7 +55,7 @@ module.exports = function(element, bpmnFactory, options, translate) {
 
       var event = (listener.get('event')) ? listener.get('event') : '<empty>';
 
-      var label = (event || '*') + ' : ' + (LISTENER_TYPE_LABEL[listenerType] || '');
+      var label = (event || '*') + ' : ' + translate(LISTENER_TYPE_LABEL[listenerType] || '');
```
barmac: Have you tested this? I think it might be difficult to get the actual template string if we use a variable here. Let's keep the previous, explicit behaviour, shall we?

lib/provider/camunda/parts/ListenerDetailProps.js
```diff
@@ -138,7 +138,7 @@ module.exports = function(group, element, bpmnFactory, options, translate) {
           listener = getSelectedListener(element, node),
           listenerType = ImplementationTypeHelper.getImplementationType(listener);
 
-      value.listenerValueLabel = LISTENER_TYPE_LABEL[listenerType] || '';
+      value.listenerValueLabel = translate(LISTENER_TYPE_LABEL[listenerType] || '');
```
barmac: I'd rather not invoke `translate` with variables as it might be difficult to extract the template for localization.

barmac:
First and foremost thank you for your pull request! :) I left a few comments in the code. Please make sure that you adhere to our commit guidlines, which are linked in the [CONTRIBUTING.md file](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/.github/CONTRIBUTING.md#creating-a-pull-request).

lib/provider/camunda/parts/implementation/Listener.js
rumata28: Sure, I did.

lib/provider/camunda/parts/ListenerDetailProps.js
rumata28: ok

rumata28: changed it.

lib/provider/camunda/parts/implementation/Listener.js
rumata28: changed it.

rumata28:
Thank you for review of this PR.

I made a new commit with changes, which address (I hope) your remarks.

pinussilvestrus:
Hi @rumata28,
it would be very cool if you can squash your commits and clean up the commit history with the style @barmac posted above. As a result, we should have fewer commits which make it clear, what the pull request is supposed to add or to fix. If you need advice, don't be afraid to ask :)

rumata28:
Squashed and force-pushed it back to the branch.  Hope I did not screw it up. :)

pinussilvestrus:
Thanks! Just one more thing: Can you modify your commit message to fit our [guideline](https://www.conventionalcommits.org/)? It would be something like

`feat(providers): improve overall localization`

You could even consider splitting the commit into 2 or 3 since (as you described above, cf. https://github.com/bpmn-io/bpmn-js-properties-panel/pull/295#issue-261257415) there are different things achieved inside your pull request.

However, @barmac can you have a look at whether the made changes fit your requests?

rumata28:
Changed commit message as per recommendation of Niklas Kiefer.
This is still one commit, all changes are dedicated to improving overall localization.

If you see that it is not ok, could you please be more specific how exactly you would like to split it into several?

lib/provider/camunda/parts/implementation/Listener.js
```diff
@@ -55,7 +55,7 @@ module.exports = function(element, bpmnFactory, options, translate) {
 
       var event = (listener.get('event')) ? listener.get('event') : '<empty>';
 
-      var label = (event || '*') + ' : ' + (LISTENER_TYPE_LABEL[listenerType] || '');
+      var label = (event || '*') + ' : ' + LISTENER_TYPE_LABEL[listenerType] || '';
```
barmac: This should remain the same, cf.
```javascript
label = (undefined || '*') + ' : ' + undefined || '';
"* : undefined"

label = (undefined || '*') + ' : ' + (undefined || '');
"* : "
```

```suggestion
      var label = (event || '*') + ' : ' + (LISTENER_TYPE_LABEL[listenerType] || '');	
```

lib/provider/camunda/parts/VariableMappingProps.js
```diff
@@ -233,7 +233,12 @@ module.exports = function(group, element, bpmnFactory, translate) {
   group.entries.push(entryFactory.selectBox({
     id: 'variableMapping-inOutType',
     label: translate('Type'),
-    selectOptions: inOutTypeOptions,
+    selectOptions: inOutTypeOptions.map(function(option) {
```
barmac: Options should be translated as soon as they appear in code as strings. Putting translate function in map will make it difficult actually translate applications.

barmac:
I've added a few more comments. Let's fix this, as we're close to merge :)

lib/provider/camunda/parts/implementation/Listener.js
rumata28: done

lib/provider/camunda/parts/VariableMappingProps.js
rumata28: Agreed.
Done.

rumata28:
Resolved both findings.

Also, found some 2 more options list - localized them too (those are execution and task listener event types).

rumata28:
I added one more commit dedicated to translation of element template.

rumata28:
Guys, do you have any idea on when (or if) you plan to publish it?

barmac:
Sure, it's now published in [v0.30.0](https://github.com/bpmn-io/bpmn-js-properties-panel/releases/tag/v0.30.0).

rumata28:
Thank you!

PR #305, opened 2019-06-11
fixed Ids require

chewcw:
Fixes # Seems like a bug when instantiating Ids, the Ids constructor is not found.

### DIFF
--- a/lib/Utils.js
+++ b/lib/Utils.js
@@ -5,7 +5,7 @@ var domQuery = require('min-dom').query,
     is = require('bpmn-js/lib/util/ModelUtil').is,
     forEach = require('lodash/forEach'),
     domify = require('min-dom').domify,
-    Ids = require('ids').default;
+    Ids = require('ids');
 
 var SPACE_REGEX = /\s/;
 


nikku:
In which scenario is this a problem?

chewcw:
The problem arise when i was trying to click add form field in a user task.
image

philippfromme:
What is your setup?

chewcw:
> What is your setup?

I was using the sample application code below and ```npm run all```, then serve the dist directory.
[https://github.com/bpmn-io/bpmn-js-examples/tree/master/properties-panel](https://github.com/bpmn-io/bpmn-js-examples/tree/master/properties-panel)

philippfromme:
I cannot reproduce your issue. Bootstrapping the app and working with forms in the properties panel works fine for me.

Have you changed the code of the example?

philippfromme:
I can reproduce your issue. The example broke after the last update of the [bpmn-js-properties-panel](https://github.com/bpmn-io/bpmn-js-properties-panel/) dependency. I'm looking into it.

nikku:
I updated the example. This should not be a problem, anymore.

nikku:
Cf. https://github.com/bpmn-io/bpmn-js-examples/tree/master/properties-panel

PR #307, opened 2019-06-14
296 sanitize entities when building html

### LINKED ISSUE #296: Sanitize entities when building HTML
#### Description

We do not properly sanitize entities (labels, ...) we build the properties panel UI from. This leads to [UI bugs](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/286#issuecomment-469174622) (cf. screenshot below) or script injection in the worst case.

An example for missing entity escaping can be found [here](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/lib/factory/SelectEntryFactory.js#L51).

#### Expected Behavior 

* [ ] We properly sanitize user input in all places where it is being used to build the properties panel.
* [ ] Escaping is [ported to all our properties panels](https://github.com/bpmn-io/bpmn-js-properties-panel/issues/296#issuecomment-476146350)
* [ ] DMN https://github.com/bpmn-io/dmn-js-properties-panel/issues/9
* [x] CMMN https://github.com/bpmn-io/cmmn-js-properties-panel/issues/63

#### Additional Context

This is how the UI looks like if a description such as `List<Foo>` is rendered: 

image

barmac:
Together with default escaping provided in diagram-js this should fix the issue.

Closes #296


barmac:
Still, there is work to be done on dmn-js and cmmn-js counterparts.

lib/factory/EntryFieldDescription.js
```diff
@@ -11,11 +28,18 @@ var MARKDOWN_LINK_REGEX = /\[([^)]+)\]\(([^\]]+)\)/g;
  * @return {String}
  */
 function linkify(text) {
-  return text.replace(MARKDOWN_LINK_REGEX, '<a href="$2" target="_blank">$1</a>');
-}
+  return text.replace(/\[([^)]+)\]\((https?:\/\/[^\]]+)\)/g, function(_, text, link) {
```
barmac: RegEx updated to link only links instead of `<a href="javascript:alert(1)">link</a>`.

barmac:
@nikku Can you have a look at this now, please?

barmac:
@nikku 
Regarding (this)[https://github.com/bpmn-io/bpmn-js-properties-panel/pull/286#issuecomment-469174622], it looks good now:
Screenshot 2019-06-17 at 16 40 15

nikku:
Unfortunately this breaks [our official way of linking from descriptions](https://github.com/camunda/camunda-modeler/blob/master/resources/element-templates/samples.json).

image

Could we safely whitelist the basic `<a href="http://some-address">LinkNAME</a>` pattern?

nikku:
I do not fully understand why we are escaping / unescaping in which order here.

What I'd be O.K. with is if we got clear and simple test coverage for the various scenarios. I've attempted to add that via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/307/commits/a8e8d3578f5a46b0ff17e0e55aa9b086fa69fb09.

Let's have a talk tomorrow and you run me through why we are doing what :confused:.

nikku:
Added tokenizing description parser: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/307/commits/8fb43572d2ad06c9b47c3fbe03321d4b7d49d14a

Issue #221, opened 2017-01-05
Possible to define i/o mappings for gateways although not supported by engine


meyerdan:
Currently it is possible to define i/o mappings for gateways although these are not supported by the engine.

Expected Behavior:
- it is not possible anymore to define i/o mappings for gateways (all of them)

nikku:
Closed via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/314.

PR #314, opened 2019-08-13
fix(helper): disallow inputOutput on bpmn:Gateway

nikku:
The Input Output tab should not be shown for Gateways, it does not make sense from the engine perspective. 

Closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/221


pinussilvestrus:
Will close https://github.com/bpmn-io/bpmn-js-properties-panel/issues/221 + https://github.com/camunda/camunda-modeler/issues/491

PR #319, opened 2019-10-14
Handle line breaks in EntryFieldDescriptions

### LINKED ISSUE #315: Line breaks in EntryFieldDescriptions are not working in v0.31.0 or later
__Describe the Bug__

After implementation of #296 (Included in release v0.31.0) it's no longer possible to use line breaks (&lt;br&gt; tag) in EntryFieldDescription. 

__Expected Behavior__

`EntryFactory#textField(options)` with
`options.description = "first line<br>second line"`
produces two line description: 
```
    first line 
    second line
```

__Actual Behavior__

Because of html character sanitization produces 
`first line<br>second line`
image

__Steps to reproduce the Behavior__

Use `EntryFactory#textField(options)` with `options.description = "first line<br>second line"`

__Tasks__

* [ ] `<br>` and `<br\s/>` yields newlines

oguzeroglu:
Closes #315

### DIFF
--- a/lib/factory/EntryFieldDescription.js
+++ b/lib/factory/EntryFieldDescription.js
@@ -5,20 +5,20 @@ var escapeHTML = require('../Utils').escapeHTML;
 /**
  * Create a linkified and HTML escaped entry field description.
  *
- * As a special feature, this description may contain both markdown
- * and plain <a href> links.
+ * As a special feature, this description may contain both markdown,
+ * plain <a href> links and <br />
  *
  * @param {String} description
  */
 module.exports = function entryFieldDescription(description) {
 
   // we tokenize the description to extract text, HTML and markdown links
-  // text and links are handled seperately
+  // text, links and new lines are handled seperately
 
   var escaped = [];
 
   // match markdown [{TEXT}]({URL}) and HTML links <a href="{URL}">{TEXT}</a>
-  var pattern = /(?:\[([^\]]+)\]\((https?:\/\/[^"<>\]]+)\))|(?:<a href="(https?:\/\/[^"<>]+)">([^<]*)<\/a>)/gi;
+  var pattern = /(?:\[([^\]]+)\]\((https?:\/\/[^)]+)\))|(?:<a href="(https?:\/\/[^"]+)">(.+?(?=<\/))<\/a>)/gi;
 
   var index = 0;
   var match;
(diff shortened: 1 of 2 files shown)


lib/factory/EntryFieldDescription.js
```diff
@@ -5,45 +5,69 @@ var escapeHTML = require('../Utils').escapeHTML;
...
 
   // we tokenize the description to extract text, HTML and markdown links
-  // text and links are handled seperately
+  // text, links and new lines are handled seperately
 
   var escaped = [];
 
   // match markdown [{TEXT}]({URL}) and HTML links <a href="{URL}">{TEXT}</a>
-  var pattern = /(?:\[([^\]]+)\]\((https?:\/\/[^"<>\]]+)\))|(?:<a href="(https?:\/\/[^"<>]+)">([^<]*)<\/a>)/gi;
+  var urlPattern = /(?:\[([^\]]+)\]\((https?:\/\/[^"<>\]]+)\))|(?:<a href="(https?:\/\/[^"<>]+)">(.+?(?=<\/))<\/a>)/gi;
```
oguzeroglu: Changed here in order to handle line breaks inside text fields of <a> tag.

[^<]*  accepts strings until < is seen, which does not go well for something like: <a href="..."> hello <br/> world </a>

So I changed this to -> .+?(?=<\/) which accepts strings until </ is seen.

barmac:
For me it looks OK if we assume that no new HTML tags will be added anytime soon.

@philippfromme?

test/spec/factory/EntryFieldDescriptionSpec.js
```diff
@@ -129,6 +129,62 @@ describe('factory/EntryFieldDescription', function() {
...
+          '<a href="https://test.com" target="_blank"> HELLO <br /> WORLD </a>' +
+        '</div>'
+      );
+    });
+
+
+    it('should ignore <a> if br used within URL field', function() {
+
+      // when
+      var html = EntryFieldDescription('<a href="https://test<br />website.com"> HELLO <br/> WORLD </a>');
```
nikku: Is this what we want? I'd rather URL-escape the BR tag in that case.

oguzeroglu: Here the thing is that our URL detector pattern does not accept URL's with **<** or **>** in it due to:
`<a href="(https?:\/\/[^"<>]+)">`

So we actually handle those as plain text, so the <br /> just adds a new line. 

Independent from the BR tag, this also does not work:
`<a href="https://www.<.com>"> SDFDSF </a>`

nikku: O.k. can we make it work then?

oguzeroglu: So If I got your point correctly what we want is that:

<a href="https://test<br />site.com"> TEST </a>

Should be considered as a valid URL, but when we click on TEST, we'd go to `test<br />site.com`?

nikku: Yes, this is what I'd consider predictable behavior. Is it hard to implement? If so, I'm happy if we stick to the current behavior for now.

oguzeroglu: I integrated this behavior into <a> tags but did not change the current behavior for markdown as the existing `< or >` characters causes problems while being converted into <a> element and there was an existing test making sure that markdowns with HTML chars should be ignored: 

> should ignore special HTML chars in markdown link

nikku:
Please port this improvement to our [dmn-js-properties-panel](https://github.com/bpmn-io/dmn-js-properties-panel) and [cmmn-js-properties-panel](https://github.com/bpmn-io/cmmn-js-properties-panel), too.

oguzeroglu:
Squashed your commit into mine. I'll do separate PRs for dmn and cmmn.

PR #339, opened 2020-07-10
Implement new colors

### LINKED ISSUE #338: Align colors with Camunda Modeler
__What should we do?__

We should implement the colors described here: https://github.com/camunda/camunda-modeler/issues/1844

__Why should we do it?__

It looks good.

oguzeroglu:
Closes #338

### DIFF
--- a/styles/_mixins.less
+++ b/styles/_mixins.less
@@ -34,7 +34,6 @@
   height: 23px;
   width: 24px;
   overflow: hidden;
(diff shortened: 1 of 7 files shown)


oguzeroglu:
We'll tackle the native UI related colors in select components with a separate ticket since they need a rework on the component itself. This is ready to get merged IMO.

PR #395, opened 2020-11-03
deps: remove bpmn-font dependency

nikku:
This removes the `bpmn-font` dependency that is not needed anymore, as the required assets are inlined / shipped with bpmn-js.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,3 @@ node_js: 14
 
 script:
   - npm run all
-before_script:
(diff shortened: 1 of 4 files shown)


nikku:
As I found out after hours of debugging this won't fly due to our wiredeps post install step.

Dependencies that are being installed of GitHub do not have the packed assets (i.e. the `dist` folder), thus the CI test failures that I could not feel out.

azeghers (approved):

PR #401, opened 2020-11-30
2007 inside connector script fix

MaxTru:

Closes [camunda/camunda-modeler#2007](https://github.com/camunda/camunda-modeler/issues/2007) (fix-upstream)

__Acceptance Criteria__

__Definition of Done__

### DIFF
--- a/test/spec/provider/camunda/ConnectorSpec.js
+++ b/test/spec/provider/camunda/ConnectorSpec.js
@@ -24,7 +24,6 @@ var domQuery = require('min-dom').query,
 var find = require('lodash/find');
 
 
-
 describe('connector', function() {
 
   var diagramXML = require('./Connector.bpmn');
@@ -858,7 +857,7 @@ describe('connector', function() {
   });
 
 
-  describe('change parameter name', function(propertiesPanel) {
+  describe('change parameter name', function() {
 
     var parameterNameInput,
         parameter;
--- a/test/spec/provider/camunda/InputOutputParameterTextSpec.js
+++ b/test/spec/provider/camunda/InputOutputParameterTextSpec.js
@@ -680,6 +680,95 @@ describe('input-output-parameterType-text', function() {
 
   });
 
+
+  describe('change parameter script resource value given external resource script format', function() {
+
+    var parameterScriptResourceValueInput,
+        parameter;
+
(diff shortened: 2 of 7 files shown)


test/spec/provider/camunda/ConnectorSpec.js
```diff
@@ -1083,7 +1082,7 @@ function getOutputParameterGroupLabel(container) {
 }
 
 function getParameterNameInput(container) {
-  return domQuery('input[id*="parameterName"]', getInputParametersGroup(container));
+  return domQuery('input[name="name"]', getInputParametersGroup(container));
```
pinussilvestrus, line 1085: Why is this change needed?

test/spec/provider/camunda/InputOutputParameterTextSpec.js
```diff
@@ -727,6 +816,10 @@ function getCursorPosition() {
   return range.startOffset;
 }
 
+function getParameterScriptValueInput(container, id) {
+  return domQuery('input[id="input-parameter-' + id + 'cam-script-resource-val"]', getInputParametersGroup(container));
```
pinussilvestrus, line 820: So it's basically not the "script value", but the "script resource value"? Maybe we should make it clear then, also [in the test itself](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/401/files#diff-3f0f09702496b6cf60a8e76b7a70db2b75efd7aa92c265f073e70a944e4c84d9R700), to not get confused by the other "script value", namely the Inline Script.

`getParameterScriptResourceValue`

pinussilvestrus (commented):
Works well 👏   just two minor thoughts to consider.

test/spec/provider/camunda/ConnectorSpec.js
MaxTru: There is no factual reason. I worked on this and played with this and accidentally rewrote it. But the result is the same. I will revert the change so to not cause confusion.

test/spec/provider/camunda/InputOutputParameterTextSpec.js
MaxTru: True, will do

pinussilvestrus (approved):

PR #405, opened 2020-12-15
fix(process-variables): prevent using variable names as identifier

pinussilvestrus:
* We used the variable name as an identifier for the items inside the process variable overview
* These can include unescaped content

--> Prevent this by using generic ids + escape variable name whenever used

Related to camunda/camunda-modeler#2031

__Acceptance Criteria__

__Definition of Done__

### DIFF
--- a/lib/provider/camunda/parts/implementation/ProcessVariables.js
+++ b/lib/provider/camunda/parts/implementation/ProcessVariables.js
@@ -11,6 +11,8 @@ var getBusinessObject = require('bpmn-js/lib/util/ModelUtil').getBusinessObject,
     is = require('bpmn-js/lib/util/ModelUtil').is,
     isAny = require('bpmn-js/lib/features/modeling/util/ModelingUtil').isAny;
 
+var escapeHTML = require('../../../../Utils').escapeHTML;
+
 var factory = require('../../../../factory/EntryFactory');
 
 var entryFieldDescription = require('../../../../factory/EntryFieldDescription');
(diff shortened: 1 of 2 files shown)


nikku:
Good stuff :clap:

lib/provider/camunda/parts/implementation/ProcessVariables.js
```diff
@@ -32,13 +34,13 @@ module.exports = function(element, translate) {
       // title ///////////////////
 
       var collapsible = factory.collapsible({
-        id: scopePrefix + name + '-collapsible',
-        title: name,
+        id: scopePrefix + 'variable- ' + idx + '-collapsible',
```
nikku, line 37: This fixes another, unrelated thing I guess? Or is it just for testing purposes?

pinussilvestrus: This actually fixes the original issue. Instead of unescaping the `name`, which would also fix the issue, I found it better to simply use generic ids. Variable names can be very long and cryptic, those ids will be easier to use and maintain.

We actually do the same in other parts, e.g. [input & output parameters](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/lib/provider/camunda/parts/implementation/OutputParameters.js#L155).

nikku: Makes sense, thanks for clarifying.

nikku (approved):

PR #412, opened 2021-01-11
Add 'participant-' prefix to name and id inputs for clarity

### LINKED ISSUE #327: Clarify the Id for Pools and plain processes

For beginners it is confusing that the process id is handled in different fields depending if you modeled a process with a pool or without.

If you model a process without a pool, you have a Id and a Name to fill:
grafik

If you model a process inside a pool, you have to fill a Process Id and a Process Name:
grafik

Rename the labels: 

* Process Id and Process Name for the first picture.
* Participant Id, Participant Name, Process Id and Process Name for the second picture.

Keep it as it is.

__Additional context__

This will help beginners in trainings a lot, as it is confusing that the same process needs different fields if there is pool or not. The behaviour of the engine and the Cockpit is easier to understand.

azeghers:

Closes #327

__Acceptance Criteria__

__Definition of Done__

### DIFF
--- a/lib/provider/bpmn/parts/IdProps.js
+++ b/lib/provider/bpmn/parts/IdProps.js
@@ -15,7 +15,7 @@ module.exports = function(group, element, translate, options) {
   // Id
   group.entries.push(entryFactory.validationAwareTextField(translate, {
     id: options.id || 'id',
-    label: translate('Id'),
+    label: options.label || translate('Id'),
     description: description && translate(description),
     modelProperty: 'id',
     getProperty: function(element) {
--- a/lib/provider/bpmn/BpmnPropertiesProvider.js
+++ b/lib/provider/bpmn/BpmnPropertiesProvider.js
@@ -1,7 +1,8 @@
 'use strict';
 
 
-var inherits = require('inherits');
+var inherits = require('inherits'),
+    is = require('bpmn-js/lib/util/ModelUtil').is;
 
 var PropertiesActivator = require('../../PropertiesActivator');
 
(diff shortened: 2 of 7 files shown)


azeghers:
@andreasgeier 

This is how the properties panel looks for a participant :

image

And this is how it looks for any other element :

image

lib/provider/bpmn/parts/IdProps.js
```diff
@@ -15,7 +15,7 @@ module.exports = function(group, element, translate, options) {
   // Id
   group.entries.push(entryFactory.validationAwareTextField(translate, {
     id: options.id || 'id',
-    label: translate('Id'),
+    label: options.label || translate('Id'),
```
nikku, line 18: What about translating here, as we already translate the description as part of the `IdProps`, too.

andreasgeier (approved):
Looks as expected

lib/provider/bpmn/parts/IdProps.js
azeghers: The `option.label` attribute is defined in the `getIdOptions` of the properties provider as such :
```javascript
    return { id: 'participant-id', label: translate('Participant Id') };
```
So it already should be translated by the time it ends up in this function.
 (I don't know how to link to a file inside of the PR)

azeghers: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/412/files#diff-9578e9e6d7be08c860d340b59799ea827673c391d9163b522c8199e7e9a5515fR135-R140

nikku: Got it. So the assumption is: If something is passed, it is always translated already.

nikku: However, we handle the `description` props differently.

These are translated, after the have been passed, cf. https://github.com/bpmn-io/bpmn-js-properties-panel/blob/e9f0346513f333d8804ab822a9a1a615835e531a/lib/provider/bpmn/parts/IdProps.js#L13.

nikku: My proposal would be: We translate consistently, inside `*Props`, always.

nikku: Alternative: Consistently translate _outside_ the `*Props`, always.

PR #427, opened 2021-02-01
Improve UI existing event definitions

### LINKED ISSUE #421: Improve UI of existing eventDefinitions
__What should we do?__

The currently implemented UI for eventDefinitions can be confusing. Goal of this issue is to refactor and improve the existing UI. The goal is not to completly rethink / rework it (this will be done as part of major properties panel overhaul).

The UI shall be in line with the new camunda:errorEventDefinitions for external serviceTasks. 

In particular this affects:

- [x] errorEventDefinition
- [x] messageEventDefinition
- [x] signalEventDefinition
- [x] escalationEventDefinition

We shall:
* Update labels
* In some cases, add validation
* Change value/label of selectBox/dropdown when selecting a global entity.

See:
Screenshot_20210125_101116
Screenshot_20210125_101130
Screenshot_20210125_101201
Screenshot_20210125_101213

__Why should we do it?__

See overall objective https://github.com/camunda/camunda-modeler/issues/2070

azeghers:
This PR edits labels and adds validation to a few event definition inputs in the properties panel.

Closes #421 

__Acceptance Criteria__

__Definition of Done__


MaxTru:
Screenshot_20210201_153803


The UI looks a bit too "full" with the descriptions in my opinion.

I would propose:
* Let`s get rid of the word `global` it does IMO not add value.
* Instead of `Variable`, let's call it `Process Variable` (because that is how we also name it in I/O Mapping Component)
* I would just omit the "description" and the `of the global xyz`. I think we dont have an appropriate approach to put this info in the given properties panel concept. Reason
  * The way it is implemented now, it is too heavy
  * And alternative would be to put `... (of referenced Error)` behind the label. But this would look cluttered

@andreasgeier WDYT?

nikku:
Proposal: Descriptions are always full sentences.

Proposal: `Variable to store...` :arrow_right: `Escalataion Code Variable` + human readable description.

Regarding "too full", do you mean adding descriptions make it look "too full"? There is a number of things that could be done (and will be done), some day. One thing is consistent spacing / visual grouping (elements, their labels, with or without description). I'd not sacrifice the descriptions.

Other than that I'm :+1: for all suggestions.

test/spec/provider/bpmn/EscalationEventDefinitionSpec.js
```diff
@@ -296,4 +301,101 @@ describe('escalation-event-properties', function() {
 
   });
 
+  describe('escalation code variable', function() {
+
+    it('should fetch escalation code variable of an escalation event definition element', inject(function(propertiesPanel, selection, elementRegistry) {
```
MaxTru, line 306: naming: sometimes you call the test case `escalation code variable`, sometimes just `variable`

lib/provider/camunda/parts/implementation/ErrorEventDefinition.js
```diff
@@ -36,29 +35,52 @@ module.exports = function(
       id: 'error-element-message',
       label: translate('Error Message'),
       referenceProperty: 'errorRef',
-      modelProperty: 'errorMessage'
+      modelProperty: 'errorMessage',
+      description: translate('of the global Error')
     })
   );
 
   if (showErrorCodeVariable) {
```
MaxTru, line 43: I believe test cases (should show / not show / set / get / validate) for errorCodeVariable and errorMessageVariable are missing

test/spec/provider/bpmn/EscalationEventDefinitionSpec.js
```diff
@@ -296,4 +301,101 @@ describe('escalation-event-properties', function() {
 
   });
 
+  describe('escalation code variable', function() {
```
MaxTru, line 304: Test case for the added validation feature is missing

MaxTru (changes requested):
Works for me, see some definition of done comments

nikku:
Whatever you guys agree upon is fine for me, too.

lib/Utils.js
```diff
@@ -111,8 +111,9 @@ function refreshOptionsModel(businessObject, referencedType) {
   var model = [];
   var referableObjects = findRootElementsByType(businessObject, referencedType);
   forEach(referableObjects, function(obj) {
+    var errorCodeLabel = obj.errorCode ? ' (code=' + obj.errorCode + ')' : '';
```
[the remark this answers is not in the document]
azeghers: Sorry I misunderstood the scope of this function, I'm updating it.

lib/Utils.js
```diff
@@ -103,6 +103,20 @@ function addEmptyParameter(list) {
 
 module.exports.addEmptyParameter = addEmptyParameter;
```
MaxTru, line 105:
```suggestion


```

lib/Utils.js
```diff
@@ -103,6 +103,20 @@ function addEmptyParameter(list) {
...
+  var label = obj.name || '';
+
+  if (obj.errorCode)
+    label += ' (code=' + obj.errorCode + ')';
+  if (obj.escalationCode)
+    label += ' (code=' + obj.escalationCode + ')';
+
+  return label;
+}
+
```
MaxTru, line 119:
```suggestion

```

lib/provider/camunda/parts/implementation/ErrorEventDefinition.js
```diff
@@ -2,7 +2,9 @@
 
 var entryFactory = require('../../../../factory/EntryFactory'),
     cmdHelper = require('../../../../helper/CmdHelper'),
-    elementReferenceProperty = require('../../../bpmn/parts/implementation/ElementReferenceProperty');
+    elementReferenceProperty = require('../../../bpmn/parts/implementation/ElementReferenceProperty'),
+    utils = require('../../../../Utils');
+
```
MaxTru, line 7:
```suggestion

```

lib/provider/camunda/parts/implementation/ErrorEventDefinition.js
```diff
@@ -13,52 +15,73 @@ module.exports = function(
   var getValue = function(modelProperty) {
```
MaxTru, line 15: In ErrorEventDefiniton.js we use getValue and setValue helpers, we don't do that in EscalationEventDefinitions.js.

Is there a reason for that (it is in principle doing the same thing, isn't it)? If not, we should do it the same way in both files.

MaxTru (commented):
Looks good to me!

I esp. like the test coverage now.

Some final comments, and then it can be merged from my perspective.

Please send screenshot to @andreasgeier so that he can also approve (or comment).

lib/provider/camunda/parts/implementation/ErrorEventDefinition.js
azeghers: The only thing is that the escalation one is single use, this one is called twice.
The other examples of validation aware text fields don't use generic getters and setters, so this one is sort of the odd one.
I can align it regardless if you think it's better.

MaxTru: well when ErrorEventDefinitions is the only one, where we call them twice, that is a IMO a valid reason why we have separate helper functions for getting and setting :+1:

andreasgeier (approved):
✅  Greenlight*

*In the absence of feasible alternatives, I agree with the current solution. Ideally, we should not need description texts that stay forever in the UI and we should have a way to better structure such a form set. I'll consider this issue for the property panel redesign.

MaxTru (approved):

PR #432, opened 2021-02-22
fix(element-templates): always log id and name with validation error

MaxTru:

fix upstream camunda/camunda-modeler#2111
fix upstream camunda/camunda-modeler#2038 (duplication by accident)
__Acceptance Criteria__

__Definition of Done__

### DIFF
--- a/test/spec/provider/camunda/element-templates/ElementTemplatesLoaderSpec.js
+++ b/test/spec/provider/camunda/element-templates/ElementTemplatesLoaderSpec.js
@@ -190,8 +190,8 @@ describe('element-templates - ElementTemplatesLoader', function() {
           var errors = e.errors;
 
           expect(messages(errors)).to.eql([
-            'template id <foo> already used',
-            'template id <foo> already used'
+            'template(id: <foo>, name: <Foo>): template id <foo> already used',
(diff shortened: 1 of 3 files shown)


test/spec/provider/camunda/element-templates/ElementTemplatesLoaderSpec.js
```diff
@@ -188,10 +188,10 @@ describe('element-templates - ElementTemplatesLoader', function() {
         var errorListener = sinon.spy(function(e) {
 
           var errors = e.errors;
-
+          console.log(messages(errors));
```
pinussilvestrus, line 191: I think this can be removed :)

pinussilvestrus (commented):
One little comment. Besides, I pretty much like the better logging 👍 

I bother a bit that we now pipe the `template` through all validation methods, but I guess there is no real opportunity to improve it - especially if we reduce the validator by using the JSON schema directly, some day.

test/spec/provider/camunda/element-templates/ElementTemplatesLoaderSpec.js
MaxTru: Ups :-)

PR #438, opened 2021-03-01
Support multiple tab providers

nikku:
This implements a tab provider mechanism similar to the one we bpmn-js uses for palette and context pad.

---

Required by https://github.com/camunda/camunda-modeler/issues/1992.

### DIFF
--- a/lib/PropertiesPanel.js
+++ b/lib/PropertiesPanel.js
@@ -37,6 +37,7 @@ var getBusinessObject = require('bpmn-js/lib/util/ModelUtil').getBusinessObject;
 var HIDE_CLASS = 'bpp-hidden';
 var DEBOUNCE_DELAY = 300;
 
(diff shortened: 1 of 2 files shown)


pinussilvestrus (commented):
I really like that 👏 

I created an example plugin to verify it works with the Camunda Modeler: https://github.com/pinussilvestrus/properties-provider-plugin-example. Super smooth to update the tabs 👍 

I'd go for merging it, but I'm also ok to wait for another voice to review it. The next step would then be to update the element-templates-modal plugin to consume the new API.

nikku:
Aaa, I was searching for the place the Camunda Modeler hooks into the properties panel. Thanks for the :point_right:.

nikku:
If you think it is fine, please add your :+1:. We generally practice four eyes, not many eyes :wink:

pinussilvestrus (approved):
🎸

PR #440, opened 2021-03-04
Element Templates Scopes v2

### LINKED ISSUE #423: Scoped bindings: allow "scopes" to be an Array, being backwards compatabile to current concept
__What should we do?__

In order to support the configuration of global entities (e.g., `bpmn:error` or `bpmn:message`), we will use scoped bindings (see parent issue). Since one might have multiple of these global entities, we need to adjust the scoped binding implementation. It shall be possible to define scoped bindings as Array (rather than as object). 

This issue only aims at adding the capability to define scoped bindings as Array.

**As-is situation**
```json
...
    "scopes": {
      "camunda:Connector": {
        "properties": [
          {
            "label": "ConnectorId",
            "type": "String",
            "value": "My Connector HTTP - GET",
            "binding": {
              "type": "property",
              "name": "connectorId"
            }
          },
          {
            "label": "Method",
            "type": "String",
            "value": "GET",
            "binding": {
              "type": "camunda:inputParameter",
              "name": "method"
            }
          },
          {
            "label": "URL",
            "type": "String",
            "value": "https://bpmn.io",
            "binding": {
              "type": "camunda:inputParameter",
              "name": "url"
            }
          },
          {
            "label": "Response",
            "type": "String",
            "value": "wsResponse",
            "binding": {
              "type": "camunda:outputParameter",
              "source": "${S(response)}",
              "scriptFormat": "freemarker"
            }
          }
        ]
      }
...
```

**To-be situation**
```json
...
"scopes": [
    {
      "type": "camunda:Connector",
      "properties": [
        {
          "label": "ConnectorId",
          "type": "String",
          "value": "My Connector HTTP - GET",
          "binding": {
            "type": "property",
            "name": "connectorId"
          }
        },
        {
          "label": "Method",
          "type": "String",
          "value": "GET",
          "binding": {
            "type": "camunda:inputParameter",
            "name": "method"
          }
        },
        {
          "label": "URL",
          "type": "String",
          "value": "https://bpmn.io",
          "binding": {
            "type": "camunda:inputParameter",
            "name": "url"
          }
        },
        {
          "label": "Response",
          "type": "String",
          "value": "wsResponse",
          "binding": {
            "type": "camunda:outputParameter",
            "source": "${S(response)}",
            "scriptFormat": "freemarker"
          }
        }
      ]
    }
  ]
...
```

Note that, the *to-be situation* shall be the new default (e.g., in terms of documentation), yet the old format shall be understood (but only for the existing type, which is `camunda:Conncetor`.

Tasks:

- [ ] Allows `scopes` to contain a list of objects, where the `type` property defines the type
- [ ] Implement backwards-compatabile logic so that existing element templates with scoped bindings (only type we suppor as-is is `camunda:Connector`, so only this case needs to be taken care of) are still valid
- [ ] Update [docs and samples](https://github.com/camunda/camunda-modeler/tree/develop/docs/element-templates)

__Why should we do it?__

See overall objective https://github.com/camunda/camunda-modeler/issues/2070
Overall concept https://miro.com/app/board/o9J_lZ7EwN0=/ (see right side of the board)

pinussilvestrus:
This allows the new format for scopes inside element templates.

```js
{
  "name": "My Template",
  "id": "com.example.template",
  "version": 1,
  "appliesTo": [
    "bpmn:ServiceTask"
  ],
  "properties": [],
  "scopes": [
    {
      "type": "camunda:Connector",
      "properties": [ ]
  ]
}
```

It also ensures backward compatibility for old `camunda:Connector` scopes.

```js
{
  "name": "My Template",
  "id": "com.example.template",
  "version": 1,
  "appliesTo": [
    "bpmn:ServiceTask"
  ],
  "properties": [],
  "scopes": {
    "camunda:Connector": {
      "properties": [ ]
  }
}
```

Closes #423

### DIFF
--- a/lib/provider/camunda/element-templates/util/handleLegacyScopes.js
+++ b/lib/provider/camunda/element-templates/util/handleLegacyScopes.js
@@ -1,0 +1,46 @@
+'use strict';
+
+var assign = require('min-dash').assign,
+    forEach = require('min-dash').forEach,
+    keys = require('min-dash').keys,
+    isObject = require('min-dash').isObject;
+
+/**
+ * Converts legacy scopes descriptor to newer supported array structure.
+ *
+ * For example, it transforms
+ *
+ * scopes: {
+ *   'camunda:Connector':
+ *     { properties: []
+ *   }
+ * }
+ *
+ * to
+ *
+ * scopes: [
+ *   {
+ *     type: 'camunda:Connector',
+ *     properties: []
+ *   }
+ * ]
+ *
+ * @param {ScopesDescriptor} scopes
+ *
+ * @returns {Array}
+ */
+module.exports = function handleLegacyScopes(scopes) {
+  var scopesAsArray = [];
+
+  if (!isObject(scopes)) {
+    return scopes;
+  }
+
+  forEach(keys(scopes), function(scopeName) {
+    scopesAsArray.push(assign({
+      type: scopeName
+    }, scopes[scopeName]));
+  });
+
+  return scopesAsArray;
+};
--- a/test/spec/provider/camunda/element-templates/util/handleLegacyScopesSpec.js
+++ b/test/spec/provider/camunda/element-templates/util/handleLegacyScopesSpec.js
@@ -1,0 +1,47 @@
+'use strict';
+
+var isArray = require('min-dash').isArray;
+
+var handleLegacyScopes = require('lib/provider/camunda/element-templates/util/handleLegacyScopes');
+
+
+describe('element-templates/util - handleLegacyScopes', function() {
+
+  it('should transform legacy scopes descriptor', function() {
+
+    // given
+    var templates = require('../fixtures/scopes-single-connector');
+
+    var scopesDescriptor = templates[0].scopes;
+
+    // when
+    var scopes = handleLegacyScopes(scopesDescriptor);
+
+    // then
+    expect(isArray(scopes)).to.be.true;
+
+    expect(scopes).to.have.length(1);
+
+    expect(scopes[0].type).to.eql('camunda:Connector');
+
+    expect(scopes[0].properties).to.eql(scopesDescriptor['camunda:Connector'].properties);
+  });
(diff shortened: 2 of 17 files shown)


lib/provider/camunda/element-templates/util/handleLegacyScopes.js
```diff
@@ -0,0 +1,46 @@
...
+
+var assign = require('min-dash').assign,
+    forEach = require('min-dash').forEach,
+    keys = require('min-dash').keys,
+    isObject = require('min-dash').isObject;
+
+/**
+ * Converts legacy scopes descriptor to newer supported array structure.
+ *
+ * For example, it transform
```
MaxTru, line 11:
```suggestion
 * For example, it transforms
```

test/spec/provider/camunda/element-templates/util/handleLegacyScopesSpec.js
```diff
@@ -0,0 +1,45 @@
...
+
+    // when
+    var scopes = handleLegacyScopes(scopesDescriptor);
+
+    // then
+    expect(isArray(scopes)).to.be.true;
+
+    expect(scopes).to.have.length(1);
+
+    expect(scopes[0].type).to.eql('camunda:Connector');
```
MaxTru, line 25: Should also check that the properties stay the same?

```suggestion
    expect(scopes[0].type).to.eql('camunda:Connector');
    expect(scopes[0].properties).to.eql(scopesDescriptor['camunda:Connector'].properties);
```

MaxTru (commented):
Very nice! Just two minor things, afterwards good to fly :rocket:

lib/provider/camunda/element-templates/util/handleLegacyScopes.js
pinussilvestrus: Thanks!

MaxTru (approved):

PR #447, opened 2021-03-09
Return an empty array instead of undefined when querying empty extension elements

azeghers:
This PR cleans up the `getExtensionElements` helper, making it return `[]` instead of `undefined` for empty extension elements, to avoid unnecessary checks, and updates every usage of it in the properties panel.

This will be a breaking change for next version, for anyone importing it, namely zeebe.

### DIFF
--- a/lib/helper/ExtensionElementsHelper.js
+++ b/lib/helper/ExtensionElementsHelper.js
@@ -7,23 +7,20 @@ var is = require('bpmn-js/lib/util/ModelUtil').is;
 
 var ExtensionElementsHelper = {};
 
-var getExtensionElements = function(bo) {
-  return bo.get('extensionElements');
-};
-
 ExtensionElementsHelper.getExtensionElements = function(bo, type) {
-  var extensionElements = getExtensionElements(bo);
+  var elements = [];
+  var extensionElements = bo.get('extensionElements');
+
   if (typeof extensionElements !== 'undefined') {
     var extensionValues = extensionElements.get('values');
     if (typeof extensionValues !== 'undefined') {
-      var elements = extensionValues.filter(function(value) {
+      elements = extensionValues.filter(function(value) {
         return is(value, type);
       });
-      if (elements.length) {
-        return elements;
-      }
     }
   }
+
+  return elements;
 };
 
 ExtensionElementsHelper.addEntry = function(bo, element, entry, bpmnFactory) {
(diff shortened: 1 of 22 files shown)


lib/helper/ExtensionElementsHelper.js
```diff
@@ -7,23 +7,20 @@ var is = require('bpmn-js/lib/util/ModelUtil').is;
 
 var ExtensionElementsHelper = {};
 
-var getExtensionElements = function(bo) {
```
azeghers, line 10: This helper inside the helper was confusing, since its name is the same as one of the exports, and it does something else. It was a one-liner anyway, and was only used in one place, so I removed it.

philippfromme:
Can you explain why this is a breaking change and how we should handle it?

azeghers:
Anyone specifically expecting an `undefined` return value and checking against it would get a truthy value from `[]`.
But you're right in asking how to proceed next because I didn't think it through...

azeghers:
@philippfromme Coming back to the topic after aligning with Max,
There is an assumption that very few external users are actually utilizing the `extensionElementsHelper` or would be affected by a breaking change there.
Hence, the only thing to adjust is the Zeebe properties panel, as well as documenting the change in the changelog if someone is indeed affected by the refactoring.
Let me know what you think about it.

pinussilvestrus:
Do we already know how much we would have to touch inside the `camunda-bpmn-js` Zeebe Properties Provider?

azeghers:
About a dozen files use this helper, half of them being tests :
image

pinussilvestrus:
My gut feeling says it's great that we are already good on track with updating our tools when using this helper function (cf. https://github.com/camunda/camunda-bpmn-js/issues/19). _However_ I'd like to at least communicate this as breaking change (in commit + changelog for the next release). 

We don't know [how many plugins](https://github.com/camunda/camunda-modeler-process-io-specification-plugin/blob/7a3f9d5b1040b30e4374c49dd85e585ecf74b74a/client/properties-panel/props/process-io-props.js#L37) are actually using it, we don't want to update them on our own, so lets at least communicate it.

azeghers:
My intention was to add it to the changelog regardless.
About the [plugin](https://github.com/camunda/camunda-modeler-process-io-specification-plugin/blob/7a3f9d5b1040b30e4374c49dd85e585ecf74b74a/client/properties-panel/props/process-io-props.js#L37) you mention, I chose not to update it because its last change dates back two years and its usage of the helper wouldn't be affected by the new return value anyway. After that I can't find any more usages or rewrites of the helper that would need updating.
Does that make sense?

pinussilvestrus:
That's fine. Let's make sure we don't only add it to the Changelog, but also to the commit.

nikku (approved):
:clap:

PR #451, opened 2021-03-22
fix(collapsible): display necessary part of title

barmac:
https://user-images.githubusercontent.com/28307541/112017124-9e445680-8b2d-11eb-865d-b640e3ee04e1.mp4

Related to https://github.com/camunda/camunda-modeler/issues/2166

### DIFF
--- a/styles/properties.less
+++ b/styles/properties.less
@@ -422,8 +422,11 @@ label.bpp-hidden {
 
   .bpp-collapsible__title {
     display: block;
+    flex: 1 0;
+    max-width: min-content;
     margin: 0;
     padding-right: 6px;
+
     white-space: nowrap;
     overflow: hidden;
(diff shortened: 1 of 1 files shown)


styles/properties.less
```diff
@@ -112,7 +118,7 @@
     flex-direction: row;
     align-items: center;
 
-    > .bpp-error-message {
+    >.bpp-error-message {
```
nikku, line 121: What and why do we reformat the whole file?

barmac: lol I committed only what I wanted to remove 🙈

barmac: It should be OK now. Thanks for the review!

barmac:
I've removed the unintended formatting and added a small cleanup on top.

nikku (approved):

PR #452, opened 2021-03-22
WIP chore(test): test removal of templates on root level

MaxTru:

related to camunda/camunda-modeler#2128
requires bpmn-io/diagram-js#540

### DIFF
--- a/lib/provider/camunda/element-templates/parts/DescriptionProps.js
+++ b/lib/provider/camunda/element-templates/parts/DescriptionProps.js
@@ -86,14 +86,10 @@ module.exports = function(
             var type = businessObject.$type,
                 eventDefinitionType = getEventDefinitionType(businessObject);
 
-            var newElement = replace.replaceElement(element, {
+            replace.replaceElement(element, {
               type: type,
               eventDefinitionType: eventDefinitionType
             });
-
-            if (!is(newElement, 'bpmn:Process') && !is(newElement, 'bpmn:Collaboration')) {
-              selection.select(newElement);
-            }
           }
         }
       ]
--- a/lib/provider/camunda/element-templates/parts/DescriptionProps.js
+++ b/lib/provider/camunda/element-templates/parts/DescriptionProps.js
@@ -7,7 +7,8 @@ var domClasses = require('min-dom').classes,
     domify = require('min-dom').domify,
     domQuery = require('min-dom').query;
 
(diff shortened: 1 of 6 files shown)


MaxTru:
WIP: will work as soon as we have requires bpmn-io/diagram-js#540

lib/provider/camunda/element-templates/parts/DescriptionProps.js
```diff
@@ -90,7 +91,9 @@ module.exports = function(
               eventDefinitionType: eventDefinitionType
             });
 
-            selection.select(newElement);
+            if (!is(newElement, 'bpmn:Process') && !is(newElement, 'bpmn:Collaboration')) {
```
nikku, line 94: Why is this check necessary? What would we like to achieve? 

General remark: 

* From my point of view, this behavior in general could be part of the [`SelectionBehavior`](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/selection/SelectionBehavior.js) and not burried here, too.

test/spec/provider/camunda/element-templates/parts/DescriptionPropsSpec.js
```diff
@@ -180,109 +246,153 @@ describe('element-templates/parts - Description Properties', function() {
...
+
+        // then
+        var collaboration = elementRegistry.get('Collaboration_1'),
+            root = canvas.getRootElement(),
+            rootBo = getBusinessObject(root);
+
+        expect(collaboration).not.to.exist;
+
+        expect(root).to.exist;
+        expect(is(root, 'bpmn:Collaboration')).to.be.true;
```
nikku, line 391: There is a few more things that I'd like to see specified, not here but in bpmn-js most likely:

* Does replace keep the old ID? 
* Does root replace keep existing message and error event definitions?
* Does replace keep existing, unknown extension elements?

lib/provider/camunda/element-templates/parts/DescriptionProps.js
MaxTru: good point, see https://github.com/bpmn-io/diagram-js/pull/540/commits/33acbcf9cac50173a4487328580b58489ca2cf80

MaxTru:
See https://github.com/bpmn-io/diagram-js/pull/540#issuecomment-815639997

PR #453, opened 2021-03-23
Remove error event definitions when changing service task implementation type

### LINKED ISSUE #441: Remove camunda:ErrorEventDefinition-related properties on changing Service Task implementation type
Only external service tasks should be able to reference `camunda:ErrorEventDefinitions`, but changing the implementation type currently doesn't remove error-related attributes.

Following #435

azeghers:
This PR adds a condition that removes all `camunda:ErrorEventDefinition` extensions when changing the implementation type of a service task from external to any other.

Closes #441


nikku (approved):
Looks good. 

Thanks :clap:

PR #455, opened 2021-03-25
feat(element-templates): validate templates via JSON Schema

### LINKED ISSUE #386: Validate element templates and handle errors accordingly.
> **This is an epic issue which may be broken down into multiple tasks, if needed**

__What should we do?__

Currently there is only a rudimentary validation of element templates implemented (ie. see [validator.js](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/lib/provider/camunda/element-templates/Validator.js)). Therefore wrong configurations will potentially lead to technical error message in the console.

Therefore this epic shall include:

- [ ] Use https://github.com/camunda/element-templates-json-schema, https://github.com/bpmn-io/json-schema-validator when parsing element templates. This will allow a more extensive validation of templates
- [ ] Ignore invalid templates and inform user accordingly
- [ ] To be decided: do not sync invalid templates from Cawemo (in fact this would probably better be done on Cawemo side - just mentioning it for completness)
- [ ] To be decided (depends on https://github.com/bpmn-io/element-templates-json-schema/issues/5): apply template validation using different schema version based on element template version

(*note that these questions might change based on kickoff / clarification of this issue*)

__Why should we do it?__

See above.

__Context__

Blocked by https://github.com/bpmn-io/element-templates-json-schema/issues/5

pinussilvestrus:
This (partly) replaces the existing `PropertiesPanel#Validator` functionality with our `element-templates-json-schema`. This means that we now
* Validate `property.type` <-> `property.binding.type` combinations (Note that this needs an update of [our documentation](https://github.com/camunda/camunda-modeler/tree/develop/docs/element-templates#bindings), since beforehand, we did not strictly enforce those combinations, but informed possible errors)
* Keep the validation of versioning and schema compatibility inside the original Validator
* Move all other validation to the Schema itself

I decided to keep the existing test cases to have another verification, although most of them only test the existing JSON Schema.

Bildschirmfoto 2021-03-25 um 13 09 24

----

Requires https://github.com/camunda/element-templates-json-schema/pull/27 to be merged + released + integrated
Closes #386
Related to https://github.com/camunda/camunda-modeler/issues/1939
Related to https://github.com/camunda/camunda-modeler/issues/2159

### DIFF
--- a/lib/provider/camunda/element-templates/Validator.js
+++ b/lib/provider/camunda/element-templates/Validator.js
@@ -1,43 +1,14 @@
 'use strict';
 
 var isArray = require('min-dash').isArray,
-    isObject = require('min-dash').isObject,
-    keys = require('min-dash').keys;
+    filter = require('min-dash').filter;
 
 var semver = require('semver');
 
-var handleLegacyScopes = require('./util/handleLegacyScopes');
+var SchemaValidator = require('@bpmn-io/json-schema-validator').Validator;
 
-var DROPDOWN_TYPE = 'Dropdown';
-
-var VALID_TYPES = [ 'String', 'Text', 'Boolean', 'Hidden', DROPDOWN_TYPE ];
-
-var PROPERTY_TYPE = 'property',
-    CAMUNDA_PROPERTY_TYPE = 'camunda:property',
-    CAMUNDA_INPUT_PARAMETER_TYPE = 'camunda:inputParameter',
-    CAMUNDA_OUTPUT_PARAMETER_TYPE = 'camunda:outputParameter',
-    CAMUNDA_IN_TYPE = 'camunda:in',
-    CAMUNDA_OUT_TYPE = 'camunda:out',
-    CAMUNDA_IN_BUSINESS_KEY_TYPE = 'camunda:in:businessKey',
-    CAMUNDA_EXECUTION_LISTENER = 'camunda:executionListener',
-    CAMUNDA_FIELD = 'camunda:field',
-    CAMUNDA_CONNECTOR = 'camunda:Connector',
-    CAMUNDA_ERROR_EVENT_DEFINITION = 'camunda:errorEventDefinition';
-
-var VALID_BINDING_TYPES = [
-  PROPERTY_TYPE,
-  CAMUNDA_PROPERTY_TYPE,
-  CAMUNDA_INPUT_PARAMETER_TYPE,
-  CAMUNDA_OUTPUT_PARAMETER_TYPE,
-  CAMUNDA_IN_TYPE,
-  CAMUNDA_OUT_TYPE,
-  CAMUNDA_IN_BUSINESS_KEY_TYPE,
-  CAMUNDA_EXECUTION_LISTENER,
-  CAMUNDA_FIELD,
-  CAMUNDA_ERROR_EVENT_DEFINITION
-];
-
-var SUPPORTED_SCHEMA_VERSION = require('@camunda/element-templates-json-schema/package.json').version;
+var SUPPORTED_SCHEMA_VERSION = require('@camunda/element-templates-json-schema/package.json').version,
+    schema = require('@camunda/element-templates-json-schema/resources/schema.json');
 
 
 /**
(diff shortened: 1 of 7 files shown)


test/spec/provider/camunda/element-templates/parts/CustomProps.json
```diff
@@ -591,7 +591,7 @@
       },
       {
         "value": "println end",
-        "type": "String",
+        "type": "Hidden",
```
nikku, line 594: What is this? A simple template fix?

test/spec/provider/camunda/element-templates/util/validateSpec.js
```diff
@@ -8,13 +8,13 @@ describe('element-templates/util - validate', function() {
...
     // given
-    var templateDescriptors = require('../fixtures/error-property-invalid');
+    var templateDescriptors = require('../fixtures/error-bindings-invalid');
 
     // when
     var errors = validate(templateDescriptors);
 
     // then
-    expect(errors).to.have.length(2);
+    expect(errors).to.have.length(7);
```
nikku, line 17: :+1:

lib/provider/camunda/element-templates/Validator.js
```diff
@@ -405,3 +172,28 @@ function getSchemaVersion(schemaUri) {
...
+ *
+ * @param {Array} schemaErrors
+ *
+ * @return {Array}
+ */
+function filteredSchemaErrors(schemaErrors) {
+  return filter(schemaErrors, function(err) {
+
+    // (1) regular errors are customized from the schema
+    if (err.keyword === 'errorMessage') {
```
nikku, line 187: What are "irrelevant" errors that we silently ignore?

test/spec/provider/camunda/element-templates/parts/CustomProps.json
pinussilvestrus: The template was actually invalid. `camunda:executionListener` bindings can't have the property type `String`.

Cf. https://github.com/camunda/camunda-modeler/tree/develop/docs/element-templates#camundaexecutionlistener

lib/provider/camunda/element-templates/Validator.js
pinussilvestrus: These are basically errors produced by the JSON Schema which live _alongside_ the primary errors why the validation failed.

As an example, let's have a look [at this example](https://github.com/camunda/element-templates-json-schema/blob/master/test/fixtures/invalid-camunda-in-type.js#L44-L71).

We define rules which have to fail since a given condition is fulfilled. The schema produces one (or more) let's say "primary" errors which indicate this rule. However, the JSON Schema is producing more errors that are not relevant for the user to see.

* One sub-schema (definition) which was expected to succeed is not green --> all the other sub-schemas are also not fulfilled (e.g. the template itself is no array)
* `oneOf` is not fulfilled for the rules
* The `if` clause is not fulfilled 
* ....

This brings us to the assumption: all relevant errors rather have a custom error message defined by us (which is the case currently) or are simple data type errors.

pinussilvestrus: I hope that clarifies it a bit, we can add more comments to make it clearer if we want to.

nikku: Makes sense. Would be beneficial to have a summary of what you said in the code I guess. I'd otherwise struggle to understand what this is about, next time I check.

nikku (approved):
All good from my side.

pinussilvestrus:
Thanks for the review ❤️ 
I will handle the getting-done once we finish the upstream PR (https://github.com/camunda/element-templates-json-schema/pull/27)

lib/provider/camunda/element-templates/Validator.js
pinussilvestrus: Will add 👍

pinussilvestrus:
I also tried out whether we want to display the `dataPath` as additional error information.

Bildschirmfoto 2021-03-26 um 08 45 11

my few cents: It does add much more extra value, the error message (template name, id, clear error message) should be clear enough.

pinussilvestrus:
Oh, `merge-me` was a bit too quick maybe. However, you already approved it 💁🏻‍♂️

I added another issue to consider the `dataPath`: https://github.com/bpmn-io/bpmn-js-properties-panel/issues/459

nikku:
^^ All good from `merge-me` perspective :wink:

PR #460, opened 2021-03-29
Fix for opening and closing of properties, in order to work in web co…

### LINKED ISSUE #458: Input parameters not working inside a web component
__Describe the Bug__

The input parameters cannot be opened inside a webcomponent, because 'min-dom' is using document and not document.body.<web component>.shadowDom

The error occurs in InputOutputParameter.js:77 because entryNode is null:
```
  result.setOpen = function(value) {
    var entryNode = domQuery('[data-entry="' + collapsible.id + '"]');
    collapsible.setOpen(value, entryNode);
  };
```
__Steps to Reproduce__

1. Build the properties panel inside a web component and add 2 input parameters. 
2. Try to open one of the input parameters from the icon 'arrow' on the left side.
3. Error is shown in the console

__Expected Behavior__

The input parameters can be edited inside a web component.

__Environment__

- Host (Browser/Node version), if applicable: [e.g. MS Edge 18, Chrome 69, Node 10 LTS]
Chrome and Edge Chromium
- OS: [e.g. Windows 7]
Windows 10
- Library version: [e.g. 2.0.0]
0.40.0

simeon-penev:
…mponents
Closes #458

### DIFF
--- a/lib/provider/camunda/parts/implementation/Errors.js
+++ b/lib/provider/camunda/parts/implementation/Errors.js
@@ -153,7 +153,7 @@ function getErrorsEntries(element, bpmnFactory, options, translate) {
           return;
         }
 
-        entry.setOpen(false);
+        entry.setOpen(false, entryNode.parentElement);
       });
 
     }
--- a/lib/provider/camunda/parts/implementation/InputOutputParameter.js
+++ b/lib/provider/camunda/parts/implementation/InputOutputParameter.js
@@ -74,8 +74,8 @@ module.exports = function(parameter, bpmnFactory, options, translate) {
 
   var isOpen = options.isOpen || collapsible.isOpen;
 
-  result.setOpen = function(value) {
-    var entryNode = domQuery('[data-entry="' + collapsible.id + '"]');
+  result.setOpen = function(value, parentElement) {
+    var entryNode = domQuery('[data-entry="' + collapsible.id + '"]', parentElement);
     collapsible.setOpen(value, entryNode);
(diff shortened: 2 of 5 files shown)


barmac:
Thanks for your contribution. Can you please update the commit according to [the Contribution Guidelines](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/.github/CONTRIBUTING.md#creating-a-pull-request)? Also, we'd love to have a test case that prevents regression.

barmac (changes requested):
Please see the comment above.

simeon-penev:
We are working on the test

simeon-penev:
@barmac We provided the test in the pull request. Can you please squash-merge?

barmac:
Thanks for the update. Before we merge it, I'd love to verify that it actually works within the shadow DOM. Can you add a test which verifies it?

holkwen:
@barmac Thank you for your comment. We've checked your existing test infrastructure, and unfortunately it does not seem to support the shadow DOM. In particular, the used TestContainer is not suitable for the shadow DOM, as stated in its docs. The correctness of the solution can be checked logically, and we also added a test to prevent regression, but building a new test infrastructure for the tiny fix would be too much effort for us at the moment.

nikku:
I believe it is safe to merge this without the shadow DOM test. If it works for @simeon-penev and @holkwen it will work for others, too.

barmac:
Let's do that. Thank you for your contribution!

PR #462, opened 2021-04-07
fix(element-templates): use pre-compiled schema validator

pinussilvestrus:
This uses the pre-compiled validator + schema: https://github.com/bpmn-io/element-templates-validator

* no more compiling on runtime, prevents CSP errors in the Modeler distro
* reduced bundle size
* the validator now defines with schema version we're using

Related to camunda/camunda-modeler#2181


barmac (approved):
Confirmed that it works also with a production build of Camunda Modeler:

image

PR #465, opened 2021-04-12
feat: update existing documentation entries

nikku:
This PR changes our documentation behavior to more sensible defaults:

* we pick the first documentation element that has a
  textFormat=text/plain (or an empty text format)
* we update the documentation entry rather than resetting
  the whole list of entries

This ensures we play more nicely with others, that may use multiple
documentation entries for different use cases.

Related to https://github.com/camunda/camunda-modeler/issues/1682


barmac (approved):
Works as expected :rocket:

PR #468, opened 2021-04-23
Bump bpmn js 8.3.1

MaxTru:
Related to https://github.com/bpmn-io/bpmn-js-properties-panel/pull/467

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   },
   "license": "MIT",
   "devDependencies": {
-    "bpmn-js": "^8.3.0",
+    "bpmn-js": "^8.3.1",
     "bpmn-moddle": "^7.0.4",
     "camunda-bpmn-moddle": "^5.1.0",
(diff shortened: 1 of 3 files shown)


nikku:
The thing we've just talked about, right? Glad to see it is an instance of a broken test case.

nikku (approved):

package.json
```diff
@@ -64,6 +64,6 @@
     "semver": "^6.3.0"
   },
   "peerDependencies": {
-    "bpmn-js": "^3.x || ^4.x || ^5.x || ^6.x || ^7.x || ^8.x"
```
nikku, line 67: Why this change? What actually broke in terms of compatibility?

nikku (changes requested):
Too quick on the review. One question.

package.json
MaxTru: Corrected

nikku:
:clap:

PR #475, opened 2021-06-30
feat(element-templates): support specific local camunda:in|out

MaxTru:
fixes upstream camunda/camunda-modeler#2334

requires camunda/element-templates-json-schema#30

(Note that prior to releasing this, we need to bump the
element-templates-json-schema version - otherwise respective schemas
will be marked as invalid and ignored)

### DIFF
--- a/lib/provider/camunda/element-templates/CreateHelper.js
+++ b/lib/provider/camunda/element-templates/CreateHelper.js
@@ -272,31 +272,26 @@ function createCamundaInOutAttrs(binding, value) {
     if (binding.target && !binding.expression && !binding.variables) {
       properties.target = binding.target;
       properties.source = value;
-    }
 
-    if (binding.target && binding.expression === true && !binding.variables) {
+    } else if (binding.target && binding.expression === true && !binding.variables) {
       properties.target = binding.target;
       properties.sourceExpression = value;
-    }
 
-    if (binding.variables === 'local' && !binding.target && !binding.expression) {
+    } else if (binding.variables === 'local' && !binding.target && !binding.expression) {
       properties.local = true;
       properties.variables = 'all';
-    }
 
-    if (binding.variables === 'local' && binding.target && !binding.expression) {
+    } else if (binding.variables === 'local' && binding.target && !binding.expression) {
       properties.local = true;
       properties.source = value;
       properties.target = binding.target;
-    }
 
-    if (binding.variables === 'local' && binding.target && binding.expression) {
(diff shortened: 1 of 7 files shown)


lib/provider/camunda/element-templates/CreateHelper.js
```diff
@@ -268,36 +268,70 @@ function createCamundaInOutAttrs(binding, value) {
 
   var properties = {};
 
-  // camunda:in source(Expression) target
-  if (binding.target) {
-
-    properties.target = binding.target;
+  if (binding.type === 'camunda:in') {
```
pinussilvestrus, line 271: I think all these cases are exclusive? Maybe use if-else then?

MaxTru: adjusted 

:+1:

lib/provider/camunda/element-templates/CreateHelper.js
```diff
@@ -268,36 +268,60 @@ function createCamundaInOutAttrs(binding, value) {
...
+
+    } else if (binding.variables === 'local' && binding.sourceExpression && !binding.source) {
+      properties.local = true;
+      properties.sourceExpression = binding.sourceExpression;
+      properties.target = value;
+
+    } else if (binding.variables === 'local' && !binding.sourceExpression && !binding.source) {
+      properties.local = true;
+      properties.variables = 'all';
+    }
```
nikku, line 324: Quite a switch :sweat_smile:

nikku:
@pinussilvestrus Can you have another look at this today? I could otherwise take over the review, too.

pinussilvestrus:
@nikku Feel free to take it over! I am probably not able to look into this today.

nikku:
Will have a look.

lib/provider/camunda/element-templates/CreateHelper.js
```diff
@@ -268,36 +268,70 @@ function createCamundaInOutAttrs(binding, value) {
...
+
+    if (binding.variables === 'local' && binding.target && !binding.expression) {
+      properties.local = true;
+      properties.source = value;
+      properties.target = binding.target;
+    }
 
-    if (binding.variables === 'local') {
+    if (binding.variables === 'local' && binding.target && binding.expression) {
       properties.local = true;
```
nikku, line 290: Feedback: Kind of hard to see where we added the fix. Is it here?

nikku (approved):
I like that we're adding additional test coverage here.

Looks good to me implementation wise.

PR #476, opened 2021-07-09
chore(elementTemplates): refactor createHelper for camunda:in|out

MaxTru:
* This follows an hour-of-code discussion where we identified that the
  affected lines of code are hard to understand and maintain without
  the now added documentation


nikku (approved):
Great improvement. Thank you for the follow up.

PR #482, opened 2021-09-24
feat(properties-panel): gracefully handle incompatible providers

nikku:
Related to https://github.com/bpmn-io/internal-docs/issues/323


test/spec/PropertiesPanelSpec.js
```diff
@@ -686,6 +686,20 @@ describe('properties-panel', function() {
...
+
+      // given
+      var incompatibleProvider = {
+        getGroups: function() {
+          return [];
+        }
+      };
+
+      // then
+      propertiesPanel.registerProvider(incompatibleProvider);
```
MaxTru, line 700: Should we not check here that 'propertiesPanel.providersChanged' was not fired, that undefined was returned and potentially even that the error was logged?

MaxTru (changes requested):
I think we could improve the test case

test/spec/PropertiesPanelSpec.js
nikku: We should use internal `getProviders` API to verify the new provider is not part of the bundle.

I'll adjust the PR accordingly.

smbea (approved):
LGTM

PR #483, opened 2021-10-13
fix(inputs): don't use browser defaults for undo+redo

marstamm:
related to bpmn-io/bpmn-js#1493

implements https://github.com/bpmn-io/diagram-js/pull/573


nikku (approved):
Very nice. :+1: for also adding a little test.

PR #516, opened 2021-12-02
fix(zeebePropProvider): rename target group to calledElement

MaxTru:
Required by camunda/camunda-modeler#2586

Screenshot_20211202_204654


barmac (approved):

andreasgeier (approved):
I like that we are now consistent with spec and naming across both stacks.

Regarding the feedback that we only call processes for cloud: I believe the section makes that clear with the process input label, at least in opened state.

✅ &nbsp; Approved

PR #525, opened 2021-12-10
Reuse preact instance and adjust to new exports

barmac:
This PR:
* adds missing `package.json#files` field to include only the build artifacts
* removes unnecessary `index.js` file in the root
* transforms the package to re-use preact shipped with `@bpmn-io/properties-panel`
* adjusts `@bpmn-io/properties-panel` imports to the new structure introduced in the PR linked below.

Requires https://github.com/bpmn-io/properties-panel/pull/124 to be merged and released

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,8 +1,0 @@
-{
-  "plugins": [
-    [ "@babel/plugin-transform-react-jsx", {
-      "importSource": "preact",
-      "runtime": "automatic"
-    } ]
-  ]
-}
--- a/index.js
+++ b/index.js
@@ -1,1 +1,0 @@
-module.exports = require('./lib');
--- a/karma.config.js
+++ b/karma.config.js
@@ -70,7 +70,7 @@ module.exports = function(karma) {
               options: {
                 plugins: [
                   [ '@babel/plugin-transform-react-jsx', {
-                    'importSource': 'preact',
+                    'importSource': '@bpmn-io/properties-panel/preact',
                     'runtime': 'automatic'
                   } ]
                 ]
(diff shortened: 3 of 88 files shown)


test/spec/provider/zeebe/Forms.spec.js
```diff
@@ -44,7 +44,7 @@ describe('provider/zeebe - Forms', function() {
     BpmnPropertiesPanel,
     CoreModule,
     ModelingModule,
-    SelectionModule,
+    SelectEntryionModule,
```
pinussilvestrus, line 47: I think we renamed too many modules here 😅

barmac: Indeed!

barmac: Fixed via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/525/commits/abef6050fdafd92c8a7e73381a62f79f27de4d12

test/spec/provider/bpmn/CompensationProps.spec.js
```diff
@@ -14,7 +14,7 @@ import {
 } from 'min-dom';
 
 import CoreModule from 'bpmn-js/lib/core';
-import SelectionModule from 'diagram-js/lib/features/selection';
+import SelectEntryionModule from 'diagram-js/lib/features/selection';
```
pinussilvestrus, line 17: Same here. I think we can simply remove all these `SelectEntryionModule` changes.

barmac: Ok fixed

nikku (approved):
A fundamental improvement. :+1:

PR #526, opened 2021-12-14
test(camunda-platform): add test case to verify layout config

pinussilvestrus:
This adds a test case to verify our layout config feature we don't actively promote yet, but use internally. It's possible to pre-define the `open` state of certain groups with it. Example:

```js
import BpmnModeler from 'bpmn-js/lib/Modeler';

const modeler = new BpmnModeler({
  container: '#canvas',
  propertiesPanel: {
    parent: '#properties',
    layout: {
      groups: {
        'general': { open: true },
        'documentation': { open: false },
        'CamundaPlatform__UserAssignment': { open: true }
      }
    }
  }
});
```


MaxTru (approved):
:clap:

PR #529, opened 2021-12-16
rewrite distro test

barmac:
## test: remove require test 

Some of the dependencies expose only ESModules and this is causing
problems unless we instrument rollup to bundle the deps with
the library. This, however, leads to increased bundle size
which doesn't give any benefit as modern bundlers can handle
commonjs/esmodules interop.

### DIFF
--- a/test/distro/distroSpec.js
+++ b/test/distro/distroSpec.js
@@ -9,23 +9,8 @@ const DIST_DIR = path.join(__dirname, '../../dist');
 
 describe('modules', function() {
 
-  it('should expose CJS bundle', function() {
+  it('should expose CJS bundle', verifyExists('index.js'));
 
-    // given
-    const {
-      BpmnPropertiesPanelModule,
-      BpmnPropertiesProviderModule,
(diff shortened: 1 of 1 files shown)


test/distro/distroSpec.js
```diff
@@ -9,9 +9,12 @@ const DIST_DIR = path.join(__dirname, '../../dist');
 
 describe('modules', function() {
 
-  it('should expose CJS bundle', function() {
+  it.skip('should expose CJS bundle', function() {
```
nikku, line 12: Is this `skip` a left over, given that we shimmed `window`, too?

barmac: It won't work even with shimmed window but the error message refers to `import` being used in `bpmn-js`. I decided to leave it `skip`ped as I thought maybe some day we want to fix that.

nikku: Let us remove the shim all-together then.

barmac: Done.

Issue #537, opened 2022-01-13
The group label not translate 


polarloves:
in the source code,the group label has not translate.

i find that ,if i do not modify the source code ,i can not translate it !

![image](https://user-images.githubusercontent.com/11882911/149283469-fc1157fa-59ff-4b6a-8d69-4b4da0172e5a.png)
![image](https://user-images.githubusercontent.com/11882911/149283553-63be29a6-2168-4465-b974-b5ba6af329d6.png)

polarloves:
and i find that the header label can not be translated too.

polarloves:
now i can translate like this : 
![image](https://user-images.githubusercontent.com/11882911/149294278-700f63c2-b387-4607-ae29-6e2de233eea3.png)
![image](https://user-images.githubusercontent.com/11882911/149294393-773ff99b-f50e-4223-bd72-56f879ba508d.png)

polarloves:
pull request:https://github.com/bpmn-io/bpmn-js-properties-panel/pull/536

Abimostu:
@smbea sorry to bother you, but i find that the title could not be translated. Can you help me ? thanks 
![WX20221118-162707](https://user-images.githubusercontent.com/16587503/202656857-3fdd05dd-d115-468a-bd0b-cb7d6d2d86c8.png)

nikku:
> https://github.com/bpmn-io/bpmn-js-properties-panel/issues/537#issuecomment-1319694382

Please open a new issue and clearly layout your case.

PR #542, opened 2022-01-14
translate group label and header type

polarloves:
fix bug:https://github.com/bpmn-io/bpmn-js-properties-panel/issues/537

### DIFF
--- a/src/provider/bpmn/BpmnPropertiesProvider.js
+++ b/src/provider/bpmn/BpmnPropertiesProvider.js
@@ -1,5 +1,7 @@
 import { Group } from '@bpmn-io/properties-panel';
 
+import { forEach } from 'min-dash';
+
 import {
   CompensationProps,
   DocumentationProps,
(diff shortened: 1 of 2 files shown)


src/provider/bpmn/BpmnPropertiesProvider.js
```diff
@@ -206,17 +208,23 @@ function getGroups(element) {
...
+    this._translate = translate;
   }
 
   getGroups(element) {
+    const translate = this._translate;
     return (groups) => {
       groups = groups.concat(getGroups(element));
+      // translate group labels
+      forEach(groups, function(group) {
+        group.label = translate(group.label);
```
pinussilvestrus, line 222: With this one, we would fix the translation for the `bpmn` properties provider only. We have [multiple ones](https://github.com/bpmn-io/bpmn-js-properties-panel/tree/master/src/provider) for different use cases (Camunda Platform vs. Camunda Cloud specific).

Maybe we should add translation there as well?

barmac:
Instead of monkey-patching the labels, I'd suggest that we translate them explicitly like we already do in the element templates: https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/provider/element-templates/ElementTemplatesPropertiesProvider.js#L100

barmac (changes requested):
Cf. https://github.com/bpmn-io/bpmn-js-properties-panel/pull/542#issuecomment-1028718716

pinussilvestrus:
@polarloves any updates on this one?

smbea:
Since there hasn't been any activity on this pull request, I will be closing it for now.

PR #546, opened 2022-01-17
Add validation to called element `processId` in Call Activity

### LINKED ISSUE #346: Consistently refuse non QName process ids
__Describe the Bug__

See https://github.com/zeebe-io/zeebe-modeler/issues/237, raised by [@ajeans](https://github.com/ajeans)

__Steps to Reproduce__

1. Create a new BPMN
2. Add a call activity to it, with process-id "123"
3. Save the BPMN
4. Create a second BPMN
5. Define its id as "123"

__Expected Behavior__

If QName is indeed mandatory for process-ids, it should also be enforced in the details of the call activity.

__Environment__

OS: Ubuntu 20.04
Zeebe Modeler Version: 0.9.1

smbea:
Closes #346

Uses the same validation as `IdProps`, except:
* field remains non-mandatory 
* don't check if id already assigned because this is used to refer to an existing process Id

### DIFF
--- a/src/provider/zeebe/properties/TargetProps.js
+++ b/src/provider/zeebe/properties/TargetProps.js
@@ -9,7 +9,8 @@ import {
 
 import {
   getCalledElement,
-  getProcessId
+  getProcessId,
+  isIdValid
 } from '../utils/CalledElementUtil.js';
 
 import {
@@ -114,12 +115,17 @@ function TargetProcessId(props) {
     commandStack.execute('properties-panel.multi-command-executor', commands);
   };
 
+  const validate = (value) => {
+    return isIdValid(value, translate);
+  };
+
   return TextFieldEntry({
     element,
     id: 'targetProcessId',
     label: translate('Process ID'),
     getValue,
     setValue,
-    debounce
+    debounce,
+    validate
   });
 }
--- a/src/provider/zeebe/utils/CalledElementUtil.js
+++ b/src/provider/zeebe/utils/CalledElementUtil.js
@@ -6,6 +6,13 @@ import {
   getExtensionElementsList
 } from './ExtensionElementsUtil';
 
+const SPACE_REGEX = /\s/;
+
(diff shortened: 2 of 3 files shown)


pinussilvestrus:
Works well in general 👍  I'm just wondering: As far as I remember we limit the validation in the properties panel due to a bare minimum because we want to establish linting as the go-to validation.

This since is id validation and with this, we make sure our diagram is sound (don't use invalid ids), I'd agree to add it. Wdyt @MaxTru @nikku @andreasgeier?

(Sorry when this was already clarified somewhere 🙂 )

MaxTru:
> Works well in general +1 I'm just wondering: As far as I remember we limit the validation in the properties panel due to a bare minimum because we want to establish linting as the go-to validation.
> 
> This since is id validation and with this, we make sure our diagram is sound (don't use invalid ids), I'd agree to add it. Wdyt @MaxTru @nikku @andreasgeier?
> 
> (Sorry when this was already clarified somewhere slightly_smiling_face )

I would be okay with this PR (pragmatic decision, we already validate the ID). Once we have linting, we could easily unbuild.

smbea:
@nikku @andreasgeier Any thoughts?

MaxTru (changes requested):
Note that also an expression is valid. C.f. https://docs.camunda.io/docs/components/modeler/bpmn/call-activities/call-activities/#defining-the-called-process

So we would need to invest into the validation (either qname static value of expression). Hence, considering that we anyhow invest into this via linting soon, I would leave it...

smbea:
> Note that also an expression is valid. C.f. https://docs.camunda.io/docs/components/modeler/bpmn/call-activities/call-activities/#defining-the-called-process
> 
> So we would need to invest into the validation (either qname static value of expression). Hence, considering that we anyhow invest into this via linting soon, I would leave it...

Ok, makes sense. Should I close this PR then?

MaxTru:
Your call - I would recommend yes :-)

PR #549, opened 2022-01-20
Render HTML descriptions for Element Templates

pinussilvestrus:
Closes #547 
Closes #375 

This uses the same `preact-markup` and sanitization setup as we already use in `form-js`. I checked out the Sanitizer and believe we can safely use the rules as we have with our Markdown support in form-js, I don't see a need for specific rules in properties panel land. We already sanitize pretty carefully over there.

After a chat with @MaxTru I decided to keep it simple for now and only bring back the [HTML support for element template field descriptions](https://github.com/camunda/camunda-modeler/blob/develop/docs/element-templates/README.md#defining-template-properties). The introduction of `preact-markup` would allow us to do even crazier things as a core feature, e.g. using markdown / HTML to provide my custom properties - instead of the existing JSX syntax.

This can be easily achieved if we decide on it in the future.

<image src="https://user-images.githubusercontent.com/9433996/150322182-3702bece-c968-44b7-8522-738746267933.png" height="500px" />

PR #551, opened 2022-01-21
fix(element-templates): do not render empty descriptions

pinussilvestrus:
Follow up of #549 

> I've found a problem within the templates example. There's a description with `undefined` as the content.
> 
> image

_Originally posted by @barmac in https://github.com/bpmn-io/bpmn-js-properties-panel/issues/549#issuecomment-1017721031_


MaxTru (approved):
:wave:

PR #558, opened 2022-01-25
docs: correct usage in README

nikku:
We did use the old import until now.

The one documented works at `bpmn-js-properties-panel@next`.


nikku:
What we should also fix / add is concrete install instructions.

As of today (`master`) it is required to install `@bpmn-io/properties-panel` as a peer dependency.

barmac (approved):

PR #555, opened 2022-01-24
Sprint 1 - C8 Templates

pinussilvestrus:
Closes #540 
Depends on https://github.com/camunda-cloud/zeebe-bpmn-moddle/pull/20

What's in
- [x] add basic `cloud-element-templates` feature, based on existing C7 implementation
- [x] add commands for supported bindings
  - [x] `zeebe:taskDefinition:type`
  - [x] `zeebe:taskHeader`
  - [x] `zeebe:input`
  - [x] `zeebe:output`
- [x] add custom props for supported bindings
- [x] proper default types (e.g. `zeebe:taskDefinition:type`  => `hidden`?) ==> _to be clarified_

What's not in
- add `optional` property (cf. [#559](https://github.com/bpmn-io/bpmn-js-properties-panel/issues/559)) 
- JSON schema validation
- scopes support
- make base element templates more reusable (namespace agnostic)
  - `ElementTemplatesGroup`
  - `getTemplateId` helpers

image

Issue #523, opened 2021-12-06
As properties provider, I can easily move properties / groups to a specific position


pinussilvestrus:
__Is your feature request related to a problem? Please describe__

Currently, the order of a group is (mostly) defined by the order it got added in the properties provider, e.g. [`CamundaPlatformPropertiesProvider#_getGroups`](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/next/src/provider/camunda-platform/CamundaPlatformPropertiesProvider.js#L45). It should be possible to move groups to a certain position (absolute, e.g. `first`, `last` - or relative, e.g. below `General`, above `Implementation`) in an easy manner.

Potentially, this is already possible by finding the group/entry and inserting it to a specific position, manually. As an example, we [add properties to the existing `General` group](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/next/src/provider/camunda-platform/CamundaPlatformPropertiesProvider.js#L148).

I could imagine that this is a returning use case, especially for plugins, to easily insert groups/entries to a certain position - not by manually iterating the collected groups, but **having helper functions** to serve these use cases.

__Describe the solution you'd like__

Providing helper functions to easily move around groups. E.g. (just sketching here)

```js
// adding a group
import { removeGroup, insertGroup, moveGroup } from 'bpmn-js-properties-provider/util';

export default class MyPropertiesProvider {

  constructor(propertiesPanel, injector) {
    propertiesPanel.registerProvider(LOW_PRIORITY, this);

    this._injector = injector;
  }

  getGroups(element) {
    return (groups) => {

      // (1) remove general group
      removeGroup(groups, 'general');

      // (2) add my group in a specific position
      insertGroup(groups, MyNewGroup, { before: 'CamundaPlatform__Implementation' });
      
      // (3) move existing group to a specific position
      moveGroup(groups, 'CamundaPlatform__UserAssignment', { after: 'CamundaPlatform__AsynchronousContinuations' });

      return groups;
    };
  }
}
```

__Describe alternatives you've considered__

Keep things as is.

__Additional context__

/

pinussilvestrus:
This was a learning for me personally after doing some ordering updates manually: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/522.

ingorichtsmeier:
As a user I asked myself: "How get the sections and properties in a section ordered?" and had a quick look at the code. But I didn't found any hint. 

[...]

PR #564, opened 2022-01-28
C8 Template groups

pinussilvestrus:
Builds on top of #555.
Related to https://github.com/camunda/camunda-modeler/issues/2673

image


pinussilvestrus:
The relevant changes for this PR are located in https://github.com/bpmn-io/bpmn-js-properties-panel/pull/564/commits/52cb56d99519c1ad73b4bff228b753536abe92e2

MaxTru (changes requested):
This is a personal request ;-) Would it be possible to have an example for custom groups in the start script for cloud-templates? It makes fuzzy testing (for this PR, but also for future changes to come) so much easier

MaxTru:
Also one thought:
Do we have an issue where we document the documentation updates that are necessary once we make this avaialble publically? Rationel I ask: I believe our [existing docs](https://github.com/camunda/camunda-modeler/tree/develop/docs/element-templates) are superb. I know we cannot update them now, since this feature is not yet supported / available public, but I would like to ensure that we document it in the used quality once we get to that point.

pinussilvestrus:
> This is a personal request ;-) Would it be possible to have an example for custom groups in the start script for cloud-templates? It makes fuzzy testing (for this PR, but also for future changes to come) so much easier

Good catch, I added it accordingly. I already had an example in fixtures I used for testing, but didn't link it to the start script 👍 

> Also one thought:
Do we have an issue where we document the documentation updates that are necessary once we make this avaialble publically? Rationel I ask: I believe our existing docs are superb. I know we cannot update them now, since this feature is not yet supported / available public, but I would like to ensure that we document it in the used quality once we get to that point.

👍  I created an issue and linked it in the epic: https://github.com/camunda/camunda-modeler/issues/2723

test/spec/provider/cloud-element-templates/fixtures/connectors.json
```diff
@@ -277,5 +277,92 @@
         }
       }
     ]
+  },
+  {
```
MaxTru, line 281: (minor) I realized that we use `connectors.json` as filename. However, this really is general C8 element templates, I guess connectors could be a bit confusing as filename isn't it?

src/provider/cloud-element-templates/properties/CustomProperties.js
```diff
@@ -72,18 +80,64 @@ export function CustomProperties(props) {
   const groups = [];
 
   const {
-    id
+    id,
+    properties,
+    groups: propertyGroups
   } = elementTemplate;
 
+  // (1) group properties by group id
```
MaxTru, line 88: Playing around with this, I realized that there is no explicit way to define the ordering (e.g., which section would appear first / which one last) of the sections. I could imagine that this is important to properly define the UX of let's say Connector (e.g., first define the URL of a REST service before adding Header fields seems natural).

Was this discussed?

MaxTru: Same applies to the `C7 Template Groups` PR of course

MaxTru (changes requested):

test/spec/provider/cloud-element-templates/fixtures/connectors.json
pinussilvestrus: Well it started with `connectors`, but it evolved 😉 I will rename the file.

src/provider/cloud-element-templates/properties/CustomProperties.js
pinussilvestrus: No, this was not discussed in particular.

The ordering is defined by the order of the `properties`, at least that is what I would expect, I didn't create a test for it, but the [`groupBy` helper](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/564/files#diff-3c1092b38e059ceff7ce76978ca0e37e08337db1acb59aeff93b8b4d6cba1f16R683) should respect that order.

So, if the first property belongs to group 1, group 1 will be shown first, and so on. The properties determine the order, as it was beforehand.

I think my intention here was to keep things simple, without coming up with complex ordering functionality inside the `groups`. If we invent something like that, we would have to make sure that the ordering is always side effect-free, something that can cause some complexity (e.g. cf. the way how [electron solves menu items ordering](https://www.electronjs.org/docs/v14-x-y/api/menu#menu-item-position))

pinussilvestrus: I also have the feeling, whatever we come up with to explicitly order template groups, that should work kind of equally (syntax-wise) to order my general properties panel groups, which we don't have yet (cf. https://github.com/bpmn-io/bpmn-js-properties-panel/issues/523).

Maybe a good follow-up task, since I believe we should handle this in an orderly manner. But I also have the sense that this would be a nice usability improvement for template creators. Maybe something we could ask for feedback (especially in the Connectors department). 👍

nikku: Not sure if I am missing the point here but the order of groups should be mandated by the following:

* Order of apperance in the template (first defined, first rendered)
* Overruled by [group definitions](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/564/files#diff-f4cf7fe9d4214e16a89074330306e719da28548aec922eefaaa4e056cb08cf92R353)

Same applies for properties defined; => order of definition = order of apperance.

MaxTru: Ups - I missed that the ordering can be defined by the order in the template :-) Then - all good!

A test case for that would be superb IMO.

pinussilvestrus: Thanks for the clarification 👍  I added a test case to verify

> Order of appearance in the template (first defined, first rendered)

If we would go for the second rule @nikku mentioned (`Overruled by group definitions`), that would _always_ apply since the `group` needs to be defined in any case to be displayed.

test/spec/provider/cloud-element-templates/fixtures/connectors.json
pinussilvestrus: Renamed via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/564/commits/5cb36c227028a2489c440a5bf82f3771f0312822

MaxTru (approved):
Great feature

PR #567, opened 2022-02-04
Add support for bpmn-js 9

marstamm:
This PR ensures that the properties panel supports drilldown.

The changes made here:
 - When the root element is a subProcess, we assume it is a collapsed subprocess
 - Process Variable discovery also works for collapsed subprocess planes

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,8 +105,8 @@
     "zeebe-bpmn-moddle": "^0.11.0"
   },
   "peerDependencies": {
-    "bpmn-js": "9.x",
-    "diagram-js": "8.x",
+    "bpmn-js": "8.x || 9.x",
+    "diagram-js": "7.x || 8.x",
     "@bpmn-io/properties-panel": "0.10.x"
   }
 }
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "@testing-library/preact": "^2.0.1",
     "babel-loader": "^8.2.2",
     "bpmn-js": "9.0.0-alpha.1",
+    "bpmn-js-legacy": "npm:bpmn-js@^8.9.1",
     "bpmn-moddle": "^7.1.2",
     "camunda-bpmn-moddle": "^6.1.1",
     "chai": "^4.3.4",
--- a/package.json
+++ b/package.json
@@ -68,14 +68,14 @@
     "@rollup/plugin-node-resolve": "^13.0.0",
     "@testing-library/preact": "^2.0.1",
     "babel-loader": "^8.2.2",
-    "bpmn-js": "^8.9.1",
+    "bpmn-js": "9.0.0-alpha.1",
     "bpmn-moddle": "^7.1.2",
     "camunda-bpmn-moddle": "^6.1.1",
     "chai": "^4.3.4",
     "copy-webpack-plugin": "^9.0.0",
     "cross-env": "^7.0.3",
     "css-loader": "^5.2.6",
-    "diagram-js": "^7.8.2",
+    "diagram-js": "^8.1.1",
     "eslint": "^7.32.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -104,8 +104,8 @@
     "zeebe-bpmn-moddle": "^0.10.0"
   },
   "peerDependencies": {
-    "bpmn-js": "8.x",
-    "diagram-js": "7.x",
+    "bpmn-js": "9.x",
```
pinussilvestrus, line 107: Just for my understanding, we need to add `bpmn-js@9` as peer dependency because of the needed `isPlane` util, right? Or is there another reason why it won't work with `bpmn-js@8` anymore?

marstamm: Yes, this is for the `isPlane` utility. I considered having a utility in this repo as well but then we have to consider changes in 2 repos if we want to make changes

pinussilvestrus: I was thinking whether we could make this backward compatible in some way, [as we did with the `isImplicit` method](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/render/BpmnPropertiesPanelRenderer.js#L172), so we could avoid a using `bpmn-js@9` in any case, and mark both versions as peer dependencies

```json
"bpmn-js": "^8.x || ^9.x"
```

Only if it's feasible of course 🙂 Wdyt?

marstamm: I'll look into it, it should absolutely be feasible from what I can see :+1:

marstamm: Added with [fbad6b9](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/567/commits/fbad6b9287a03a733f55c103271e18b3d9861bef). Can't use a `getDi` util either, because that's new in `9` as well :man_shrugging:

src/render/PanelHeaderProvider.js
```diff
@@ -134,4 +135,14 @@ function isConditionalFlow(element) {
...
 
   return businessObject.conditionExpression && is(sourceBusinessObject, 'bpmn:Activity');
+}
+
+
+// helpers //////////
+function isPlane(element) {
+
+  // Backwards compatibility for bpmn-js<8
+  var di = element && (element.di || getBusinessObject(element).di);
```
pinussilvestrus, line 145:
```suggestion
  const di = element && (element.di || getBusinessObject(element).di);
```

pinussilvestrus (commented):
Works like a charm! 👏 Only one minor thing, then we can merge it 👍

src/render/PanelHeaderProvider.js
marstamm: :facepalm: To much time spent in `bpmn-js` ES5 land

marstamm: Fixed it with [ecd47d6](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/567/commits/ecd47d661d8d365aec19eca82585da9810a5f3c4)

package.json
nikku: @marstamm Consider having a test case around to verify the bpmn-js@8 support:

Install it as bpmn-js-legacy: npm:bpmn-js@8.

Create properties panel in an integration test and see that it does not blow up.

nikku: Cf. https://stackoverflow.com/questions/56134857/how-to-install-npm-package-under-alias-or-different-name.

nikku: Also consider not adding it. People can always upgrade.

marstamm: Adding the integration tests was easy enough, so I decided there is no reason not to add them. Cf. [3696d6e](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/567/commits/3696d6e5d16a073f658f5fcd8ed80ed1ffe45aa5)

pinussilvestrus (approved):
Nice!

PR #568, opened 2022-02-07
chore(CHANGELOG): update breaking changes

barmac:


nikku (approved):

nikku:
We could consider making this an example, too (or point to a test case that showcases the interaction).

barmac:
Perhaps we could have a super simple implementation instead. E.g. open a `<select>` when the button is clicked. Then, Camunda Modeler would just override this component.

PR #574, opened 2022-02-08
refactor: include generated task forms in form select

smbea:
Closes camunda/camunda-modeler/issues/2496

https://user-images.githubusercontent.com/25825387/152951712-876cdde1-a051-4416-8591-0aa20ab3ce10.mov

### DIFF
--- a/src/provider/camunda-platform/properties/FormDataProps.js
+++ b/src/provider/camunda-platform/properties/FormDataProps.js
@@ -13,7 +13,9 @@ import {
 
 import FormField from './FormField';
 
-import { without } from 'min-dash';
+import {
+  without
+} from 'min-dash';
 
 export function FormDataProps({ element, injector }) {
   if (!isFormDataSupported(element)) {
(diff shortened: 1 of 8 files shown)


MaxTru:
The problem still exists that "Camunda Forms" could also be modeled via "Embedded or external task forms" since that would just allow to type in a formkey, isn't it?

Was this solution discussed with Ingo?

smbea:
> The problem still exists that "Camunda Forms" could also be modeled via "Embedded or external task forms" since that would just allow to type in a formkey, isn't it?
> 
> Was this solution discussed with Ingo?

This solution was discussed with both Ingo and Nico.

What I understood from what was said is that having `formRef` for Camunda Forms instead of `formKey` is better in the long run, but didn't make sense because the form binding wasn't working as expected - which was fixed by https://github.com/camunda/camunda-modeler/issues/2484. Having that, the main issue is the confusion with the generated task forms. I also checked with Nico that this was the case.

We can sync on this for further discussion, if needed.

MaxTru:
My request would be: the solution should make sense end2end incl. our guide/docs. The documentation currently states that one can use deployment or app formKey or Form Reference to deploy a camundaForm. See https://docs.camunda.org/manual/7.16/user-guide/task-forms/#camunda-forms

With the solution implemented here, one would probably not find the formKey option. Now this in general is fine with me, I just would like to ensure that Application and Documentation are aligned.

smbea:
@MaxTru This should be updated when the stable version is release, right? is there an existing issue to update the docs to include the new properties panel stuff?

MaxTru:
No, we can update right-away, it would not be published prior to April / 7.17 anyhow (docs update from version to version update)

ingorichtsmeier:
Hi @MaxTru, the platform docs aren't optimal from a user's perspective. I can provide a pull request there to clarify them.

For example, switching the sections. **formRef** should be the first option when you want to use Camunda Forms. As they can be deployed directly from the modeler, it makes the development lifecycle much easier. 

We can keep the **formKey** with Camunda forms section for backward compatibility, but put it in the back.

MaxTru:
FYI: my concern regarding docs was resolved with https://github.com/camunda/camunda-docs-manual/pull/1149. 

@smbea, please also update our screenshot automation to ensure that the new propPanel layout here is correctly represented. IMO at least this screenshot is affected: https://github.com/camunda/camunda-docs-modeler-screenshots/blob/main/lib/userGuide.js#L29

smbea:
@MaxTru There is already an issue for updating properties-panel screenshots - https://github.com/camunda/camunda-docs-modeler-screenshots/issues/18.

src/provider/camunda-platform/properties/FormDataProps.js
```diff
@@ -54,50 +56,8 @@ function addFactory({ bpmnFactory, commandStack, element }) {
...
-        context: {
-          element,
-          moddleElement: extensionElements,
-          properties: {
-            values: [ ...extensionElements.get('values'), formData ]
-          }
-        }
-      });
-    }
+    // (2) get camunda:FormData
```
MaxTru, line 59: There is a `(2)` but there is no `(1)`

src/provider/camunda-platform/properties/FormTypeProps.js
```diff
@@ -35,43 +46,101 @@ export function FormTypeProps(props) {
...
+  const commandStack = useService('commandStack');
+
+  let extensionElements = businessObject.get('extensionElements');
 
   const getValue = () => {
     if (isDefined(businessObject.get('camunda:formKey'))) {
       return 'formKey';
     } else if (isDefined(businessObject.get('camunda:formRef'))) {
       return 'formRef';
+
```
MaxTru, line 61:
```suggestion
```

MaxTru: Or add an empty line after https://github.com/bpmn-io/bpmn-js-properties-panel/pull/574/files#diff-88ce5e1f5e98eb613c3b7855f9da6b44890685dbbafb3d46c25c3f69fcd83d38R58, but let's have it consistent

MaxTru (commented):
Good stuff! Only have minor comments

smbea:
I had forgotten about this one 🤦‍♀️. I pushed the minor code style changes

MaxTru (approved):

PR #581, opened 2022-02-08
fix: remove import of non-existing API

### LINKED ISSUE #580: Incompatible with `bpmn-js@8` due to `ModelUtil#isAny is not defined`
__Describe the Bug__

[Only recently we exposed `ModelUtil#isAny`](https://github.com/bpmn-io/bpmn-js/commit/a42ba5cd90fdc438156e262680921605feba992a). It is not available for stable builds of bpmn-js yet.

Because of that installing this project and testing with `bpmn-js@8` fails for me:

image

__Steps to Reproduce__

1. `npm install bpmn-js@8 bpmn-js-properties-panel@github:bpmn-io/bpmn-js-properties-panel`
2. create an [entry point as `index.js`](https://github.com/bpmn-io/bpmn-js-properties-panel#usage)
3. `rollup index.js > index.bundled.js`

__Expected Behavior__

We do not use this brand new API yet, but instead use what is stable.

Alternatively we backport the change to bpmn-js@8.x. I think that would be a great idea.

__Environment__

- Host (Browser/Node version), if applicable: any
- OS: any
- Library version: `master`

barmac:
Closes #580

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,7 +36,7 @@ jobs:
       run: npm install ${{ matrix.integration-deps }}
     - name: Build
       env:
-        COVERAGE: ${{ matrix.integration-deps != '' }}
+        COVERAGE: ${{ matrix.integration-deps == '' }}
       run: npm run all
     - name: Upload coverage
       run: npx codecov
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,7 +9,7 @@ jobs:
         node-version: [ 14 ]
         integration-deps:
         - diagram-js@7.x bpmn-js@8.x
-        - diagram-js@8.x bpmn-js@9.0.0-alpha.2
+        - "" # as defined in package.json
 
     runs-on: ${{ matrix.os }}
 
@@ -31,7 +31,8 @@ jobs:
           ${{ runner.OS }}-
     - name: Install dependencies
       run: npm ci
-    - name: Install dependencies for integration
+    - name: Install dependencies for integration test
+      if: ${{ matrix.integration-deps != '' }}
       run: npm install ${{ matrix.integration-deps }}
     - name: Build
       env:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,6 +7,9 @@ jobs:
       matrix:
         os: [ ubuntu-latest ]
         node-version: [ 14 ]
+        integration-deps:
+        - diagram-js@7.x bpmn-js@8.x
+        - diagram-js@8.x bpmn-js@9.0.0-alpha.2
 
     runs-on: ${{ matrix.os }}
 
@@ -28,6 +31,8 @@ jobs:
           ${{ runner.OS }}-
     - name: Install dependencies
       run: npm ci
+    - name: Install dependencies for integration
+      run: npm install ${{ matrix.integration-deps }}
     - name: Build
       env:
         COVERAGE: 1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,7 +36,7 @@ jobs:
       run: npm install ${{ matrix.integration-deps }}
     - name: Build
       env:
-        COVERAGE: 1
+        COVERAGE: ${{ matrix.integration-deps != '' }}
       run: npm run all
     - name: Upload coverage
       run: npx codecov
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -35,8 +35,13 @@ jobs:
       if: ${{ matrix.integration-deps != '' }}
       run: npm install ${{ matrix.integration-deps }}
     - name: Build
+      if: ${{ matrix.integration-deps != '' }}
+      run: npm run all
+    - name: Build with coverage
+      if: ${{ matrix.integration-deps == '' }}
       env:
-        COVERAGE: ${{ matrix.integration-deps == '' }}
+        COVERAGE: 1
       run: npm run all
     - name: Upload coverage
+      if: ${{ matrix.integration-deps == '' }}
       run: npx codecov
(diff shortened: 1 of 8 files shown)


nikku:
Looks good.

What about backporting the utility to @8?

And another one: What about adding a test case that verifies we still run O.K. against bpmn-js@8. We've run into enough issues to justify that.

barmac:
> What about backporting the utility to `bpmn-js@8`?

That would be cool but we'd still depend on the most recent minor from 8.x series in that case.

> And another one: What about adding a test case that verifies we still run O.K. against bpmn-js@8. We've run into enough issues to justify that.

I'll think about it tomorrow.

barmac:
I've added integration test via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/581/commits/9de938eefb008312ca912eb8936c3d07dfa7dcd4
It seems we have yet another problem to solve: https://github.com/bpmn-io/bpmn-js-properties-panel/runs/5123108688?check_suite_focus=true#step:7:173

barmac:
@marstamm and I just had a call to discuss the breaking tests on bpmn-js@8. I will adjust the tests in this PR and add a skip for the bpmn-js@9 special case of subprocess as the root element.

.github/workflows/CI.yml
```diff
@@ -7,6 +7,9 @@ jobs:
       matrix:
         os: [ ubuntu-latest ]
         node-version: [ 14 ]
+        integration-deps:
+        - diagram-js@7.x bpmn-js@8.x
+        - diagram-js@8.x bpmn-js@9.0.0-alpha.2
```
nikku, line 12: Should we make this optional in the happy path?

I.e. only install `integration-deps` in parts of our CI build and test against everything specified in `package.json` as a happy path?

nikku (changes requested):

.github/workflows/CI.yml
barmac: Please explain what you mean by "happy path" in this context. Can you give me an example of desired outcome?

nikku: Exactly what Martin refers to, too:

https://github.com/bpmn-io/bpmn-js-properties-panel/pull/581#discussion_r802666568

There should be a configuration where we do not install these additional dependencies, but simply test things as specified in the `package.json`.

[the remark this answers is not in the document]
nikku: Skipping in the second case would be my preference, cf. https://github.com/bpmn-io/bpmn-js-properties-panel/pull/581#discussion_r802665227.

barmac:
I've added https://github.com/bpmn-io/bpmn-js-properties-panel/pull/581/commits/3a650e7f6cc8b6047bed3d3915916240e54fea41 to make sure we don't upload coverage twice.

.github/workflows/CI.yml
```diff
@@ -28,9 +31,12 @@ jobs:
           ${{ runner.OS }}-
     - name: Install dependencies
       run: npm ci
+    - name: Install dependencies for integration test
+      if: ${{ matrix.integration-deps != '' }}
+      run: npm install ${{ matrix.integration-deps }}
     - name: Build
       env:
-        COVERAGE: 1
+        COVERAGE: ${{ matrix.integration-deps != '' }}
```
marstamm, line 39: Shouldn't coverage be
```suggestion
        COVERAGE: ${{ matrix.integration-deps == '' }}
```
so we include the bpmn-js@9 only tests?

test/spec/provider/camunda-platform/ProcessVariablesProps.spec.js
```diff
@@ -367,24 +368,28 @@ describe('provider/camunda-platform - ProcessVariableProps', function() {
     }));
 
 
-    it('should display items for plane', inject(async function(elementRegistry, selection, canvas) {
+    testForBpmnJsVersion('>=9')('should display items for plane',
```
nikku, line 371: I'm personally a fan of `CHECK && it('...', ...)` to not hide what we're doing.

So to adopt to your example here:

```
bpmnJsVersion(`>=9`) && it('should display items for plane', ...
```

.github/workflows/CI.yml
barmac: Right! 😅

test/spec/provider/camunda-platform/ProcessVariablesProps.spec.js
nikku: Or, if this is a core use-case, simplify it:

```
import {
  bpmnJs9
} from '...'


bpmnJs9() && it('test what works on nine only')
```

nikku: All just personal preference. Happy to adopt what works for you.

barmac: Hmm but in that case we hide information that the test case was skipped. WDYT?

nikku: You got a point on the skip.

Then, to be less verbose why not include a simple `it9` helper?

```
import {
  it9
} from 'test-helper'


it9('do what works with bpmn-js nine+ only');
```

nikku: Conversely, consider `it10`, `it15`.

barmac:
https://github.com/bpmn-io/bpmn-js-properties-panel/commit/3a650e7f6cc8b6047bed3d3915916240e54fea41 doesn't work in action because `COVERAGE=false` is truthy 😅

test/spec/provider/camunda-platform/ProcessVariablesProps.spec.js
barmac: I shortened this to `withBpmnJs`. `it9` sounds super cryptic to me.

barmac:
@nikku @marstamm Ready for another round.

marstamm (approved):
Looks good to me

Issue #561, opened 2022-01-27
Implement JSON Schema and Validator for C8 templates


pinussilvestrus:
__What should we do?__

It should be possible to validate my C8 templates against a JSON schema, as we already do with C7 (cf. https://github.com/camunda/element-templates-json-schema + https://github.com/bpmn-io/element-templates-validator). 

* [x] Next to the basic rules, it should validate the following bindings (cf. https://github.com/camunda/element-templates-json-schema/pull/41)
  - `zeebe:taskDefinition`
  - `zeebe:taskHeader`
  - `zeebe:input`
  - `zeebe:output `
  - `groups`
  - `optional` for inputs and outputs
* [x] [Prepare `element-templates-json-schema` to provide two packages](https://github.com/camunda/element-templates-json-schema/issues/42)
  * `@camunda/element-templates-json-schema@0.7.0/resources/schema.json`
  * `@camunda/zeebe-element-templates-json-schema@0.1.0/resources/schema.json`
* [x] [Extend `element-templates-validator` to also validate C8 templates](https://github.com/bpmn-io/element-templates-validator/issues/5)
* [ ] Integrate the schema into `cloud-element-templates` provider
  * [ ] check for defined`$schema`, throw if empty or malformed
  * [ ] validate against `@camunda/zeebe-element-templates-json-schema`

What we should consider here on top
* How do we make C7 and C8 templates distinguishable? Via `platform` property? Via the existing `$schema` property?
* How many JSON schemas should we provide? A single one? One for each platform? Three with a "base" JSON schema and two extensions? 
* Do we want to make our JSON schemas more composable / easier to maintain?
* Do we need another [`element-templates-validator`](https://github.com/bpmn-io/element-templates-validator) or shall it be agnostic?

Miro board for conceptual work: https://miro.com/app/board/uXjVOPpXIIc=/

__Why should we do it?__

A proper JSON Schema validation helps a lot to build + consume valid templates.

Child of https://github.com/camunda/camunda-modeler/issues/2640

pinussilvestrus:
During investigations to create a C8 template schema out of our [existing schema](https://github.com/camunda/element-templates-json-schema/blob/master/resources/schema.json) the smartest way possible, I experimented with [json-schema-ref-parser](https://github.com/APIDevTools/json-schema-ref-parser) to make our schemas more composable.

[...]

PR #585, opened 2022-02-17
Refactor element template providers

### LINKED ISSUE #562: Consider restructuring of element templates providers
__What should we do?__

Given the introduction of C8 templates support (#540) we now have two providers that offer element templates features (`element-templates` and `cloud-element-templates`). Although the `cloud-element-templates` provider already inherits certain functionality, we should still consider making the feature engine agnostic. This means in particular

* Make template utils namespace agnostic (e.g. `getTemplateId` helpers, `ElementTemplatesGroup`)
* Make validator namespace agnostic
* Consider refactoring of base Properties provider
* Consider refactoring of base element templates service
* Consider refactoring of base command handlers 

__Why should we do it?__

Make element templates feature an engine agnostic feature and easier to adopt for future element template implementations.

pinussilvestrus:
Closes #562 

This refactors both element templates providers to use a common base (as far it makes sense). In fact, this PR

* adds a factory to create the `ElementTemplateGroup` in both worlds, with minimal customization (get, unlink, update templates)
* removes duplicated code for the `ElementTemplates` module

This _does not_ refactor, since this would need more efforts
* The command handlers
* The custom props
* The validators (this is part of #561)


pinussilvestrus:
I will rebase this on top of the latest API changes.

pinussilvestrus:
Done, ready for review now.

barmac (approved):
Works like a charm

PR #586, opened 2022-02-17
C8 templates validation

### LINKED ISSUE #561: Implement JSON Schema and Validator for C8 templates

pinussilvestrus:
Closes #561 

Implements JSON Schema validation for C8 templates.
Note that we also validate for an existing `$schema` attribute in any C8 template, that needs to reference our schema package in some way (`@camunda/zeebe-element-templates-json-schema`).

In particular it
* adds `@bpmn-io/element-templates-validator` v0.5
* adds the `$schema` attribute to existing C8 example templates
* validates C8 templates against the JSON Schema + validates the defined `$schema` attribute

----

Snapshot from the Camunda Modeler WIP integration

image

### DIFF
--- a/src/provider/cloud-element-templates/Validator.js
+++ b/src/provider/cloud-element-templates/Validator.js
@@ -1,5 +1,19 @@
-import { Validator as BaseValidator } from '../element-templates/Validator';
+import {
+  Validator as BaseValidator,
+  filteredSchemaErrors,
+  getSchemaVersion
+} from '../element-templates/Validator';
 
+import semver from 'semver';
+
+import {
+  validateZeebe as validateAgainstSchema,
+  getZeebeSchemaPackage as getTemplateSchemaPackage,
+  getZeebeSchemaVersion as getTemplateSchemaVersion
+} from '@bpmn-io/element-templates-validator';
+
+const SUPPORTED_SCHEMA_VERSION = getTemplateSchemaVersion();
+const SUPPORTED_SCHEMA_PACKAGE = getTemplateSchemaPackage();
 
 /**
  * A Camunda Cloud element template validator.
@@ -10,8 +24,6 @@ export class Validator extends BaseValidator {
   }
 
   /**
-   * TODO(pinussilvestrus): we disable JSON schema validation for now.
-   *
    * Validate given template and return error (if any).
    *
    * @param {TemplateDescriptor} template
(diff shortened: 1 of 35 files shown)


src/provider/cloud-element-templates/Validator.js
```diff
@@ -22,9 +34,34 @@ export class Validator extends BaseValidator {
     let err;
 
     const id = template.id,
-          version = template.version || '_';
+          version = template.version || '_',
+          schema = template.$schema,
+          schemaVersion = schema && getSchemaVersion(schema);
+
+    // (1) compatibility
```
MaxTru, line 41: Would it not make more sense to first do validations (2) (ie. ensure schema is defined) before checking that the version matches?

test/spec/provider/cloud-element-templates/Validator.spec.js
```diff
@@ -0,0 +1,417 @@
...
+
+      // when
+      templates.addAll(templateDescriptor);
+
+      // then
+      expect(errors(templates)).to.be.empty;
+
+      expect(valid(templates)).to.have.length(templateDescriptor.length);
+    });
+
```
MaxTru, line 241:
```suggestion


```

test/spec/provider/cloud-element-templates/fixtures/complex.json
```diff
@@ -1,5 +1,6 @@
 [
   {
+    "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
```
MaxTru, line 3: Should we add a testCase which retrieves the schema not from unpkg but from `https://www.jsdelivr.com/` or so?

MaxTru (changes requested):
Super cool! Also like the extensive test coverage.

Just small comments

pinussilvestrus:
Great suggestions, I'll follow up shortly 👍

src/provider/cloud-element-templates/Validator.js
pinussilvestrus: 👍

pinussilvestrus: Makes much more sense though 😄

test/spec/provider/cloud-element-templates/fixtures/complex.json
pinussilvestrus: Good catch, I added it accordingly 👍 

https://github.com/bpmn-io/bpmn-js-properties-panel/pull/586/files#diff-a7d860758af72e6e3dc82603905764d2c9ce5c28e93dca29dc9e5efb91fd2bf9

MaxTru (approved):

PR #602, opened 2022-03-10
Add Markers to Input fields expecting FEEL expressions

marstamm:
This PR adds FEEL icons to all input fields that accept an expression. Note that the implementation was done in https://github.com/bpmn-io/properties-panel/pull/138 and this PR just adds the required props to the input fields.

It is also possible to define a custom property to require FEEL in the configuration (related: https://github.com/camunda/element-templates-json-schema/pull/46)


package.json
```diff
@@ -60,7 +60,7 @@
   "devDependencies": {
     "@babel/core": "^7.14.3",
     "@babel/plugin-transform-react-jsx": "^7.14.3",
-    "@bpmn-io/properties-panel": "^0.11.0",
+    "@bpmn-io/properties-panel": "0.12.0",
```
MaxTru, line 63: Why not `^0.12.0`?

marstamm:
According to the validation config, it is optional :thinking: 

https://github.com/camunda-cloud/zeebe/blob/main/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/validation/ZeebeRuntimeValidators.java#L132

marstamm:
Let's go with what the Documentation says, to avoid confusing the user

src/provider/zeebe/properties/MessageProps.js
```diff
@@ -37,6 +37,11 @@ export function MessageProps(props) {
   }
 
   const entries = [
+    {
```
MaxTru, line 40: Since we add a specific messageName input (so to have the FEEL indicator), should we not also update https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/test/spec/provider/zeebe/MessageProps.spec.js

src/provider/cloud-element-templates/properties/CustomProperties.js
```diff
@@ -317,7 +319,8 @@ function TextAreaProperty(props) {
   const {
     description,
     editable,
-    label
+    label,
+    feel
```
pinussilvestrus, line 323: We don't already validate this property via JSON Schema, right? Will we bump the schema + validator in a follow-up PR?

MaxTru: Good point!

So within this PR we should also bump https://github.com/bpmn-io/element-templates-validator to a version, which will understand the feel property.

(This would requires us to release https://github.com/bpmn-io/element-templates-validator first of course...)

marstamm: I will take care of the releases tomorrow and bump it here

MaxTru: that is my understanding, yes.

pinussilvestrus: > Right, we did not bump the schma yet. I could not find a dependency for the JSON schema version, is this done through `@bpmn-io/element-templates-validator`?

Exactly, the validator is also exposing the supported Schema Version (for both schemas).

pinussilvestrus: Cf. https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/provider/cloud-element-templates/Validator.js#L13

I think it's also good practice to add some test coverage that the validator is understanding a new property correctly, e.g. what we've done previously with the new optional binding: https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/test/spec/provider/cloud-element-templates/Validator.spec.js#L379

src/provider/zeebe/properties/MessageProps.js
marstamm: Added specific tests for the message name: https://github.com/bpmn-io/bpmn-js-properties-panel/blob/662cac38f5867c2491c7821fd2ea316908396220/test/spec/provider/zeebe/MessageProps.spec.js#L73

src/provider/cloud-element-templates/properties/CustomProperties.js
marstamm: Done with [401b8e0](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/602/commits/401b8e09c21ed67416ecc89b430344a08db4cb51) and [662cac3](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/602/commits/662cac38f5867c2491c7821fd2ea316908396220) (already squashed).

Validator test was added here: https://github.com/bpmn-io/bpmn-js-properties-panel/blob/662cac38f5867c2491c7821fd2ea316908396220/test/spec/provider/cloud-element-templates/Validator.spec.js#L396-L410

MaxTru (approved):
Really really cool feature. Together with template support :rocket:

Issue #620, opened 2022-03-22
Applying a Element Template should not open any default groups


marstamm:
__Describe the Bug__

When I apply an element Template with `entriesVisible: true`, some Groups open

![Recording 2022-03-22 at 16 26 26](https://user-images.githubusercontent.com/21984219/159518054-5bb3defc-14eb-4d6c-b8bd-8cd2c6277e4c.gif)

__Steps to Reproduce__

1. `npm run start:cloud-templates`
2. select the `entries visible` template from a selected task
4. Inputs, outputs and headers open

__Expected Behavior__

The Groups stay closed

__Environment__

- Host (Browser/Node version), if applicable: [e.g. MS Edge 18, Chrome 69, Node 10 LTS]
- OS: [e.g. Windows 7]
- Library version: dev

PR #621, opened 2022-03-22
feat(element-templates): open custom groups

### LINKED ISSUE #618: Allow sections to be opened per default

In the properties panel we display sections "closed per default". This negatively impacts users in places where quick configuration is needed: For example, in the element template case I want all (or at least the mandatory inputs) to be visible to the user upon applying the template.

capture Tjkon7_optimized

My assumption is that after the first time the user has the choice to close individual sections and using our "persistent open/close mechanism" these sections will remain closed if they are chosen again, on other elements.

* [ ] We allow all sections (or a subset of sections) to be opened per default rather than closed per default upon showing it for the first time.

None.

__Additional context__

Came up in the context of connectors. Users shall be able to see all (or at least all required) fields upon applying a template to an element or creating a template-backed element.

marstamm:
Recording 2022-03-29 at 16 38 45

Try it out with `npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel#618-element-template-sections -c 'npm run start:cloud-templates'`

closes #618 

Applying a `entriesVisible` template can open other groups. This is a bug that was present before, I created https://github.com/bpmn-io/bpmn-js-properties-panel/issues/620 to track it

### DIFF
--- a/src/provider/cloud-element-templates/properties/CustomProperties.js
+++ b/src/provider/cloud-element-templates/properties/CustomProperties.js
@@ -122,7 +122,8 @@ function addCustomGroup(groups, props) {
     id,
     label,
     component: Group,
-    entries: []
+    entries: [],
+    shouldOpen: true
   };
 
   properties.forEach((property, index) => {
--- a/src/provider/element-templates/properties/CustomProperties.js
+++ b/src/provider/element-templates/properties/CustomProperties.js
@@ -161,7 +161,8 @@ function addCustomGroup(groups, props) {
     id,
     label,
     component: Group,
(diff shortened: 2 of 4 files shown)


nikku:
Please ping me once there is something I can try out, interaction wise.

marstamm:
@nikku , you can try it out with `npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel#618-element-template-sections -c 'npm run start:cloud-templates'`

nikku (approved):
Works as intended, and simple enough in terms of implementation.

Good job :clap:

PR #622, opened 2022-03-22
Semver compare

nikku:
Semver does not properly bundle with rollup as it contains cyclic dependencies:

image

This PR replaces semver with semver-compare. Also shaves off 9kB (minified) of the bundled distribution.


barmac (approved):

nikku:
Released with `v1.0.0-alpha.7`

PR #632, opened 2022-03-28
fix(element-template): copy versioned elementTemplate

### LINKED ISSUE #631: Element Template Replace behavior is broken for versioned templates
__Describe the Bug__

When I change the type of an element that has a versioned element template applied, it get's removed. The XML still contains the version number, but not the element template ID

Recording 2022-03-28 at 14 21 44

__Steps to Reproduce__

1. Create a element template with a version number and multiple `appliesTo` elements
2. Apply to a element
3. change the element type
4. The elementTemplate is removed

__Expected Behavior__

As with unversioned element templates, the element template is still applied

__Environment__

- Host (Browser/Node version), if applicable: [e.g. MS Edge 18, Chrome 69, Node 10 LTS]
- OS: [e.g. Windows 7]
- Library version: [e.g. 2.0.0]

marstamm:
Recording 2022-03-28 at 14 20 25

closes #631


src/provider/element-templates/ReplaceBehavior.js
```diff
@@ -6,37 +10,47 @@ import { is } from 'bpmn-js/lib/util/ModelUtil';
...
     const { appliesTo } = elementTemplate;
 
     const allowed = appliesTo.reduce((allowed, type) => {
-      return allowed || is(parent, type);
+      return allowed || is(newBo, type);
     }, false);
 
     if (!allowed) {
-      return false;
+      removeELementTemplate(newBo);
```
nikku, line 38: We should use the existing `updateProperties` handler here, as we don't always know if postExecuted works on a new element (does here) or not (needs to be undone) properly.

marstamm: I just found the `unlinkTemplate` utility that does exactly that :smile:

marstamm: To ensure we use the correct utility with the right prefix, I had to mirror the code for cloud templates (like we do with all element templates code)

src/provider/cloud-element-templates/ReplaceBehavior.js
```diff
@@ -0,0 +1,48 @@
...
+
+  this.postExecuted('shape.replace', function(e) {
+    var context = e.context,
+        newShape = context.newShape,
+        newBo = newShape.businessObject;
+
+    const template = newBo.modelerTemplate;
+    const version = newBo.modelerTemplateVersion;
+
+    const elementTemplate = elementTemplates.get(template, version);
```
nikku, line 24: Should we unlink only if a template was previously applied?

src/provider/element-templates/ReplaceBehavior.js
pinussilvestrus: > like we do with all element templates code

Yes, but we recently tried to make this stuff more configurable (when possible). E.g. [the templates group](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/provider/element-templates/components/ElementTemplatesGroup.js#L60).

What we could consider is that the `elementTemplates` feature has a method `unlinkTemplate` / `removeTemplate`, so that the behavior can simply re-use this, e.g. what we have with [`applyTemplate`](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/provider/element-templates/ElementTemplates.js#L123).

Not a hard requirement, but could help us in different places.

marstamm: I'd like to keep this as a `fix`, and tackle the refactoring in a follow up

pinussilvestrus: 💯 👍

nikku:
Makes sense @marstamm. Will open that separate issue.

nikku (approved):
Works as intended.

PR #645, opened 2022-03-30
Do not use `inherits`

barmac:
Usage of `inherits` was introduced without installing https://www.npmjs.com/package/inherits dependency. Thus, the Node.js internal `inherits` was used which is not being polyfilled when bundling. Instead of installing the dependency the usage of `inherits` can be avoided altogether.


philippfromme (approved):
🚀

PR #646, opened 2022-03-30
chore: ensure we use <browser> deps, if provided

nikku:
The latest release is broken due to us using inherits (that uses Node shim per default): [https://github.com/bpmn-io/bpmn-js-properties-panel/compare/v1.0.0-alpha.8...v1.0.0-alpha.9#diff-8656a2b322b6384fe4e47[…]e4056bcc4915b2d6b34d7e630R1](https://github.com/bpmn-io/bpmn-js-properties-panel/compare/v1.0.0-alpha.8...v1.0.0-alpha.9#diff-8656a2b322b6384fe4e47cfa32ed42103d62f39e4056bcc4915b2d6b34d7e630R1).

The solution to this is to tell rollup to not use the node shim, but use inherits browser version: https://github.com/bpmn-io/bpmn-js/blob/develop/rollup.config.js#L100.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -70,7 +70,13 @@ function pgl(plugins = []) {
       ]
     }),
     json(),
-    resolve(),
+    resolve({
+      mainFields: [
+        'browser',
+        'module',
+        'main'
+      ]
+    }),
     commonjs()
   ];
 }


nikku:
I did not investigate how to prevent this from happening again; I guess an integration test on the released bundle would do; we have such tests in bpmn-js, but not really [here](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/test/distro/distroSpec.js#L17).

philippfromme (approved):
Thanks for this quick fix! I will also merge https://github.com/bpmn-io/bpmn-js-properties-panel/pull/645 as it doesn't conflict with this pull request. 👍🏻 to making sure this doesn't blow up in the future.

Issue #642, opened 2022-03-29
Do not carry over element templates during replacement


nikku:
_Context: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/632, specifically https://github.com/bpmn-io/bpmn-js-properties-panel/pull/632#pullrequestreview-924627721._

__Describe the context__

In the C7 world element templates [are] ~~had unclear semantics a template could be applied to various elements "operating on top of these elements" and being~~ carried around as the underlying element type changed. Not sure how useful that feature was in the first place, but that is how it worked.

With C8 + future bpmn-js templated elements will become dedicated, first class citizens on the diagram. We call them "email task", "email connector", and they shall be thought of as dedicated entities, "templated tasks", "connector tasks" rather than something put on top of an existing task, i.e. "service task with email template applied". Cf. [mental model, Connectors Solution Architecture](https://docs.google.com/document/d/1GhSOixwZ5A7S3ajtq7KmJv7jzW9gNFulrt2TVf1pG04/edit#heading=h.w9e0ckh5vt7z)).

In this context it does not make a lot of sense to carry over the element template as I switch from one task to another:

* I switch from template task `x` to service task => I want the results to be a plain service task
* I switch from template task `x` to template task `y` => I want the result to be a template task `y` with whatever base type is specified

If we don't go this direction we may end up in a situation where I cannot [get rid of a template via replace](https://github.com/bpmn-io/bpmn-js-connectors-extension) anymore; in some contexts thought replace is the only thing we got (i.e. web-modeler or (future) no-properties modeling).

__Describe the solution you like__

We should consider removing templates during replace _or_ create a dedicated mechanism to replace, removing templates; we shall not automagically re-apply the template on the new element.

Our existing rules on keeping properties "as fit" shall continue to apply:

* Switching from a email task (built on top of a service task) to a service task keeps compatible bindings (i.e. input/output bindings)
* Switching from an email task to a user task keeps all compatible bindings
* Switching from a send task with template `A` to a service task, template `B` changes the task type, applies a new template and carries over all compatible bindings

----

_Originally posted by @nikku in https://github.com/bpmn-io/bpmn-js-properties-panel/pull/632#pullrequestreview-924627721_

nikku:
CC @marstamm @MaxTru @christian-konrad

MaxTru:
I think:

* In the **as-is** we can support a use case where:
[...]

PR #624, opened 2022-03-23
feat: add `ElementTemplates#applyTemplate` API

nikku:
This allows us to use the `elementTemplate` service to apply templates without caring about the implementation details, i.e. if it is a Platform or Cloud template.

Mid-term this will also support our "element templates everywhere" spirit (cf. [`bpmn-io/bpmn-js-connectors-extension#apply-template-usage`](https://github.com/bpmn-io/bpmn-js-connectors-extension/compare/apply-template-usage)).

---

Related to https://github.com/bpmn-io/element-template-chooser/issues/2#issuecomment-1067806835.

Required by https://github.com/bpmn-io/element-template-chooser/issues/2

PR #648, opened 2022-03-31
feat: change task type when element template is applied

### LINKED ISSUE #572: As a user I want to change a task to a different type as an element template is applied

As I model my diagram I tend to start with plain tasks rather than executable, i.e. `bpmn:ServiceTask` ones. Everything else does slow me down because I'd have to manually make each of my tasks a service task.

As I make my diagram executable I'm revising each task, making it either a service task, business rule task, ...

What I wanted to do is to apply a "mail task" template on a blank task; that template shall make the task a `bpmn:ServiceTask` before applying the template. If it was not a `camunda:ServiceTaskLike` or `zeebe:ServiceTaskLike` it was not executable by the engine.

What I have to do is to manually change the task to be a service task first, which is annoying. A unneeded indirection:

capture ze0qpr_optimized

_Camunda Modeler v5 screen capture, showcasing `ServiceTask -> REST task` transition required._

Furthermore, when searching for an element template, the service task template wouldn't even show up for blank tasks, which makes it harder to discover.

* [ ] Be able to configure a type binding for an element template; if it does not match with the current task the task is morphed into the target type _before_ the other element template properties are applied
* [ ] Given a type binding, also display these element templates on template select (maybe as a follow up action)

:arrow_right: Goal: Make it quicker to make a task executable backed by an element template.

Consider the current limitation a feature, not a bug.

__Additional context__

Came up in the context of connectors, specifically https://github.com/camunda/web-modeler/issues/524.

marstamm:
Try it out with 
```
npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel#572-element-template-change-task-type -c 'npm run start:cloud-templates'
```

Recording 2022-03-31 at 11 10 01

This PR allows you to define a `elementType` that will be enforced when a template is applied
- We use the bpmn MetaModel to define what can be changed
- Only same base types can be used for replacing (Task, Event, Gateway). Other Combinations result in invalid templates
- Works for cloud and platform

closes #572 
closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/642


package.json
```diff
@@ -62,6 +62,7 @@
   "devDependencies": {
     "@babel/core": "^7.14.3",
     "@babel/plugin-transform-react-jsx": "^7.14.3",
+    "@bpmn-io/element-template-chooser": "0.0.5",
```
pinussilvestrus, line 65: Nice 🚀  Finally we can _really_ test templates in the properties panel test setup

src/provider/cloud-element-templates/ReplaceBehavior.js
```diff
@@ -33,7 +33,15 @@ export default class ReplaceBehavior extends CommandInterceptor {
         return;
       }
 
-      const { appliesTo } = elementTemplate;
+      const { appliesTo, elementType } = elementTemplate;
+
+      if (elementType) {
+        if (!is(newShape, elementType.value)) {
```
pinussilvestrus, line 39: I wonder what happens when `elementType.value` is undefined. But that's anyway not possible since we are validating templates beforehand.

src/provider/cloud-element-templates/Validator.js
```diff
@@ -6,6 +6,8 @@ import {
 
 import semverCompare from 'semver-compare';
 
+import BpmnModdle from 'bpmn-moddle';
```
pinussilvestrus, line 9: I see this is only a _dev dependency_. Is this intended?

src/provider/cloud-element-templates/Validator.js
```diff
@@ -70,7 +72,26 @@ export class Validator extends BaseValidator {
...
     }
 
-    // (4) JSON schema compliance
+    // (4) elementType validation
+    if (template.elementType && template.appliesTo) {
+
+      const elementType = template.elementType.value,
+            appliesTo = template.appliesTo;
+
+      // (3.1) template can be applied to elementType
```
pinussilvestrus, line 81:
```suggestion
      // (4.1) template can be applied to elementType
```

😄

src/provider/cloud-element-templates/Validator.js
```diff
@@ -70,7 +72,26 @@ export class Validator extends BaseValidator {
...
+
+      // (3.2) template only applies to same type of element
+      for (const sourceType of appliesTo) {
+        if (!canMorph(elementType, sourceType)) {
+          return this._logError(`can not morph <${sourceType}> into <${elementType}>`, template);
+        }
+      }
+    }
+
+    // (5) JSON schema compliance
```
pinussilvestrus, line 94: I wonder whether we should validate first against the schema, and then do the more specific type check. This way some generic validations would happen before which we can rely on (e.g. whether `template.appliesTo` is available or not).

No hard requirement though.

src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
```diff
@@ -452,12 +457,36 @@ export default class ChangeElementTemplateHandler {
...
+
+    // determine new task type
+    const newType = newTemplate.elementType;
+
+    if (!newType) {
+      return element;
+    }
+
+    // don't replace Task that is already the correct type
+    if (is(element, newType.value)) {
```
pinussilvestrus, line 477: So we don't allow morphing ("back") into parent types? E.g.

* element has type `serviceTask`
* `template.elementType` is `task`
* do not morph element to `task`

Which makes sense due to our story I guess, just want to make sure.

src/provider/cloud-element-templates/create/TemplateElementFactory.js
```diff
@@ -62,7 +63,7 @@ export default class TemplateElementFactory {
       throw new Error('template is invalid');
     }
 
-    const type = appliesTo[0];
+    const type = (elementType && elementType.value) || appliesTo[0];
```
pinussilvestrus, line 66: I just see we could have made [`value` required for the `elementType` binding](https://github.com/camunda/element-templates-json-schema/blob/master/packages/element-templates-json-schema-shared/src/defs/base.json#L53). 

Maybe a good follow-up.

src/provider/element-templates/Validator.js
```diff
@@ -101,7 +103,26 @@ export class Validator {
...
+
+      // (3.2) template only applies to same type of element
+      for (const sourceType of appliesTo) {
+        if (!canMorph(elementType, sourceType)) {
+          return this._logError(`can not morph <${sourceType}> into <${elementType}>`, template);
+        }
+      }
+    }
+
+    // (4) JSON schema compliance
```
pinussilvestrus, line 125: I see quite the same implementation twice. Maybe we can implement it once here, e.g. as private method `_validateElementType` of the `elementTemplate` validator, and re-use it in the other (which is already extending the first one)?

test/spec/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.spec.js
```diff
@@ -640,6 +640,69 @@ describe('cloud-element-templates - ChangeElementTemplateHandler', function() {
...
+        const task = elementRegistry.get('Task_1');
+
+        // assume
+        expect(is(task, 'bpmn:ServiceTask')).to.be.true;
+
+        // when
+        changeTemplate(task, newTemplate);
+
+        // then
+        const replacedTask = elementRegistry.get('Task_1');
```
pinussilvestrus, line 662: Do we need to access it twice? It should also work if we check the type for `task`, right?

test/spec/provider/element-templates/cmd/ChangeElementTemplateHandler.spec.js
```diff
@@ -1240,6 +1241,69 @@ describe('element-templates - ChangeElementTemplateHandler', function() {
...
+        const task = elementRegistry.get('Task_1');
+
+        // assume
+
+        // when
+        changeTemplate(task, newTemplate);
+
+        // then
+        const replacedTask = elementRegistry.get('Task_1');
+        console.log(replacedTask);
```
pinussilvestrus, line 1262: remove me

test/spec/provider/element-templates/cmd/ChangeElementTemplateHandler.spec.js
```diff
@@ -1240,6 +1241,69 @@ describe('element-templates - ChangeElementTemplateHandler', function() {
...
+      const newTemplate = require('./task-template-elementType.json');
+
+      beforeEach(bootstrap(require('./task.bpmn').default, [ newTemplate ]));
+
+      it('execute', inject(function(elementRegistry) {
+
+        // given
+        const task = elementRegistry.get('Task_1');
+
+        // assume
```
pinussilvestrus, line 1255: Remove me or assume something 🙂

test/spec/provider/cloud-element-templates/Validator.spec.js
```diff
@@ -410,6 +410,80 @@ describe('provider/cloud-element-templates - Validator', function() {
...
+        expect(valid(templates)).to.be.empty;
+      });
+
+
+      it('should reject invalid replace targets', function() {
+
+        // given
+        const templates = new Validator();
+
+        const templateDescriptor = require('./fixtures/elementType-invalid-morps');
```
pinussilvestrus, line 471:
```suggestion
        const templateDescriptor = require('./fixtures/elementType-invalid-morphs');
```

pinussilvestrus (changes requested):
From quick testing I can say it works 👏 

Implementation wise we should consider some things before we merge it. 👍

package.json
marstamm: Will leave this suggestion here so I don't forget :sweat_smile: 
```suggestion
    "@bpmn-io/element-template-chooser": "^0.0.5",
```

test/spec/provider/element-templates/cmd/ChangeElementTemplateHandler.spec.js
```diff
@@ -56,11 +56,12 @@ describe('element-templates - ChangeElementTemplateHandler', function() {
     container = TestContainer.get(this);
   });
 
-  function bootstrap(diagramXML) {
+  function bootstrap(diagramXML, elementTemplates) {
```
pinussilvestrus, line 59: Do we really need that? The other test cases come out pretty well without it

test/spec/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.spec.js
```diff
@@ -640,6 +640,69 @@ describe('cloud-element-templates - ChangeElementTemplateHandler', function() {
 
     });
 
+
+    describe('update task type', function() {
```
pinussilvestrus, line 644: Shouldn't we have test cases for

* [new and old specified](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/4f0248a4d0ee27675a2202d05c8d777803f35033/test/spec/provider/element-templates/cmd/ChangeElementTemplateHandler.spec.js#L1669)
* [no new template specified](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/4f0248a4d0ee27675a2202d05c8d777803f35033/test/spec/provider/element-templates/cmd/ChangeElementTemplateHandler.spec.js#L4271)

src/provider/cloud-element-templates/Validator.js
marstamm: It should be moved to a production dependency now

src/provider/cloud-element-templates/Validator.js
pinussilvestrus: I think it's fine 👍

src/provider/cloud-element-templates/ReplaceBehavior.js
pinussilvestrus: Related: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/648#discussion_r839540764

src/provider/cloud-element-templates/Validator.js
```diff
@@ -92,4 +113,45 @@ export class Validator extends BaseValidator {
...
+
+
+/**
+ * Checks if a given type can be morphed into another type.
+ *
+ * @param {String} sourceType
+ * @param {String} targetType
+ * @returns {Boolean}
+ */
+function canMorph(sourceType, targetType) {
```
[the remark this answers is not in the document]
marstamm: Indeed, the signature is the other way around. It does not matter as you can always morph back, but it is confusing and I will change the way we call it :+1:

src/provider/element-templates/Validator.js
marstamm: I did not see that it get's extended! That makes sense, it is the same test after all

test/spec/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.spec.js
marstamm: As the task get's replaced during `bpmnReplace.replaceElement`, the reference of `task` is outdated. 

To consider: what is the return value of `commandStack.execute('propertiesPanel.zeebe.changeTemplate')`? Currently it is undefined, but maybe we can do something like `{template, element}` with the replace element + new template. WDYT?

test/spec/provider/element-templates/cmd/ChangeElementTemplateHandler.spec.js
marstamm: The problem here is that due to changing the element Template, the `replaceBehavior` is triggered which will unlink the template unless it is present and retrievable from the registry https://github.com/bpmn-io/bpmn-js-properties-panel/blob/2eb11c9c20ad7f2d4f888a22cfb82341cc2468ad/src/provider/element-templates/ReplaceBehavior.js#L29

src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
marstamm: Let's be explicit here: If I want to replace a SubProcess with an AdHocSubprocess, this would break as well.

We don't support SubProcesses for this feature yet, but we will be on the safe side if we want to include them in the future

marstamm: So: don't use the meta-model and allow morphing to general Task

test/spec/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.spec.js
pinussilvestrus: I have a bit of feeling that this could cause troubles on consumers side. Currently, when using the command, I don't have to care about the return value, I just can rely that my provided `element` will be updated.

> the reference of `task` is outdated

Is that really the case? Isn't [this line](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/648#discussion_r839580439) preventing this?

marstamm: > Is that really the case? Isn't https://github.com/bpmn-io/bpmn-js-properties-panel/pull/648#discussion_r839580439 preventing this?

Unfortunately not. It just updates the reference for the method so the attributes are applied to the new element. All references that were gotten externally (e.g. from the `elementRegistry`) still hold the old element that got replaced

marstamm: > Currently, when using the command, I don't have to care about the return value, I just can rely that my provided element will be updated.

I'm not sure if we can still support this if we have to change element types. If we can't, this would be a breaking change and we need a clear API on how to get a current reference to the element

pinussilvestrus: I think this is something we need to clarify carefully /cc @marstamm @nikku

MaxTru: For me this would be intended and also requested as outlined [here](https://github.com/bpmn-io/bpmn-js-properties-panel/issues/642). I understand it as: 

* In the past, we primarily applied element templates on top of existing element templates. So a `changeTemplate` operation would not change the element per so, but just what is applied on top of it.
* Now conceptually, we see an element template as a "custom" task. So when you apply an template, you conceptually really make it a different kind of task. So a `changeTemplate` would actually change the element per se.
* (There are edge cases like when the template has still the same targetType and so on, but what I wrote above, would be the basic conceptual architecture)

So I would say, yes, this breaks existing behavior, but this is intended and okay.

pinussilvestrus: I agree that this behavior might be expected now. However, [we have API](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/provider/element-templates/ElementTemplates.js#L133) in place that returns the element. This is public API, so we need to adapt and communicate the breaking change.

Since the API was added in alpha, I don't expect negative consequences here. We just to have to make sure we don't rely on it in our own implementations where we use the command.

* https://github.com/camunda/camunda-modeler/blob/develop/client/src/plugins/element-templates-modal/modeler/EditorActions.js#L19
* https://github.com/bpmn-io/element-template-chooser/blob/master/src/element-template-chooser/ElementTemplateChooser.js#L75

nikku: > I have a bit of feeling that this could cause troubles on consumers side. Currently, when using the command, I don't have to care about the return value, I just can rely that my provided element will be updated.

At the moment whatever our users do is fiddling with internals. We've just now created actual APIs to rely on, via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/624.

And the new API addition is built with the replace use-case in mind already.

marstamm: Thank you for the insights.

> However, [we have API](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/provider/element-templates/ElementTemplates.js#L133) in place that returns the element.

Would this be a breaking change if we return the new element instead?

nikku: Last comment: Until this point element templates was truely "experimental", and also marked as such, everywhere (cf. [docs](https://docs.camunda.io/docs/components/modeler/desktop-modeler/element-templates/about-templates/)).

With 1.0 we're getting closer with actually _stabilizing it_. The feature proves itself holisticlly, not only hidden in the properties panel.

nikku:
@marstamm I could only have a quick look on this today. If a detailed look is needed I can see what I can do tomorrow or on Monday.

test/spec/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.spec.js
```diff
@@ -640,6 +640,69 @@ describe('cloud-element-templates - ChangeElementTemplateHandler', function() {
...
+      it('execute', inject(function(elementRegistry) {
+
+        // given
+        const task = elementRegistry.get('Task_1');
+
+        // assume
+        expect(is(task, 'bpmn:ServiceTask')).to.be.true;
+
+        // when
+        changeTemplate(task, newTemplate);
```
nikku, line 659: Proposal: Move this whole thing to our new public API (cf. https://github.com/bpmn-io/bpmn-js-properties-panel/pull/624).

It also returns the new task, out of the box.

marstamm:
@nikku , I will tackle the "return new element" topic tomorrow, so nothing urgent. @pinussilvestrus already gave some great suggestions on the code, but would be great to get some feedback if it is the correct behavior for connectors as well.

nikku:
Added a few more comments, especially around the change command.

TLDR: Let us work with the newly introduced `ElementTemplate#applyTemplate` API for "applying a template" exclusively; let us encapsulate the "element can change" logic within that API call.

nikku:
Will test drive this against connectors as soon as I can.

test/spec/provider/cloud-element-templates/fixtures/complex.json
```diff
@@ -5,8 +5,11 @@
     "id": "io.camunda.connectors.RestConnector-s1",
     "description": "A generic REST service.",
     "appliesTo": [
-      "bpmn:ServiceTask"
+      "bpmn:Task"
     ],
+    "elementType": {
+      "value": "bpmn:ServiceTask"
```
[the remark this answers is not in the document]
marstamm: We had a problem with `constraints.pattern`, where we decided to go for a string only and later wanted to add an error message. To ensure extensibility in the future, we decided to do it this way and stay on the safe side

src/provider/cloud-element-templates/create/TemplateElementFactory.js
marstamm: Follow-up PR in the element templates: https://github.com/camunda/element-templates-json-schema/pull/53

marstamm: However, as `elementType` itself could still be `undefined`, we still need the check at this position

test/spec/provider/element-templates/cmd/ChangeElementTemplateHandler.spec.js
```diff
@@ -1240,6 +1241,66 @@ describe('element-templates - ChangeElementTemplateHandler', function() {
...
+
+      beforeEach(bootstrap(require('./task.bpmn').default, [ newTemplate ]));
+
+      it('execute', inject(function(elementRegistry) {
+
+        // given
+        const task = elementRegistry.get('Task_1');
+
+        // when
+        changeTemplate(task, newTemplate);
```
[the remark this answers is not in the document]
marstamm: done with [3364078](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/648/commits/3364078ec56aa663952326b6d755c36949ae51a2)

src/provider/cloud-element-templates/Validator.js
```diff
@@ -70,7 +70,14 @@ export class Validator extends BaseValidator {
       }
     }
 
-    // (4) JSON schema compliance
+    // (4) elementType validation
+    var elementTypeError = this._validateElementType(template);
```
pinussilvestrus, line 74:
```suggestion
    const elementTypeError = this._validateElementType(template);
```

src/provider/element-templates/Validator.js
```diff
@@ -101,7 +103,14 @@ export class Validator {
       }
     }
 
-    // (3) JSON schema compliance
+    // (3) elementType validation
+    var elementTypeError = this._validateElementType(template);
```
pinussilvestrus, line 107:
```suggestion
    const elementTypeError = this._validateElementType(template);
```

test/spec/provider/element-templates/cmd/ChangeElementTemplateHandler.spec.js
```diff
@@ -2951,6 +3012,7 @@ describe('element-templates - ChangeElementTemplateHandler', function() {
         var serviceTask = elementRegistry.get('ServiceTask_1');
 
         var oldTemplate = {
+          id: 'old',
```
pinussilvestrus, line 3015: Why are these ids needed now?

marstamm: With the change to `applyTemplate` in the helper, we use the internal registry for the lookup of the applied template. If both new and old templates have no ID, the key `undefined` is overridden.

We could monkey-patch the ID in the utility, but I think we should use valid templates for testing instead of doing unrelated magic in the helpers

cf. https://github.com/bpmn-io/bpmn-js-properties-panel/blob/3364078ec56aa663952326b6d755c36949ae51a2/test/spec/provider/element-templates/cmd/ChangeElementTemplateHandler.spec.js#L4486-L4490

pinussilvestrus: Nice one, thanks 👍

package.json
```diff
@@ -62,6 +63,7 @@
   "devDependencies": {
     "@babel/core": "^7.14.3",
     "@babel/plugin-transform-react-jsx": "^7.14.3",
+    "@bpmn-io/element-template-chooser": "^0.0.5",
```
pinussilvestrus, line 66: I cherry-picked this commit inside https://github.com/bpmn-io/bpmn-js-properties-panel/pull/641, so there might be merge conflicts now. Sorry for the inconvenience!

marstamm: No Problem, I have to rebase this anyway to squash my commits :)

marstamm:
Added last review hints and squashed all commits

nikku:
From what I found during initial testing, this works like a charm! 

We are aware of https://github.com/bpmn-io/bpmn-js-connectors-extension/issues/32 as a limitation, but I guess our users will verify how crucial of a limitation that is, post April 12th :wink:.

nikku:
I found one bug, not related to this PR, but to the overall "check if properties are still valid during replace" interaction:

* Given I have a template that defines a task header `<a>`
* If I change to another template that does not define task headers
* Then the task header `<a>` binding is not cleaned up

I believe this is due to our C7 legacy of "only touching what has been defined"; it does not seem to be the appropriate behavior for C8 anymore.

test/spec/provider/cloud-element-templates/Validator.spec.js
```diff
@@ -410,6 +410,80 @@ describe('provider/cloud-element-templates - Validator', function() {
...
+        // given
+        const templates = new Validator();
+
+        const templateDescriptor = require('./fixtures/elementType-invalid-morphs');
+
+        // when
+        templates.addAll(templateDescriptor);
+
+        // then
+        expect(errors(templates)).to.contain('template(id: <example.com.TaskToGateway>, name: <Element Type (Task -> Gateway)>): can not morph <bpmn:ServiceTask> into <bpmn:ExclusiveGateway>');
```
nikku, line 477: What is the reason for this limitations?

src/provider/element-templates/Validator.js
```diff
@@ -210,3 +246,45 @@ export function filteredSchemaErrors(schemaErrors) {
     return false;
   });
 }
+
+
+// helpers ///////////////////////
+
+const moddle = new BpmnModdle();
```
nikku, line 253: I see.. https://github.com/bpmn-io/bpmn-js-properties-panel/pull/648/commits/2fe461897b40b89667cd3c88e35f75b3e0835ad2#r841676679.

test/spec/provider/cloud-element-templates/Validator.spec.js
marstamm: We want to prevent the changing to broken types, like `Task -> Process` or `Task -> SequenceFlow`. As the simplest approach, we decided to only allow `Task -> Task`, `Gateway -> Gateway` and `Event -> Event`

cf. https://github.com/camunda/element-templates-json-schema/pull/48#discussion_r834478450

nikku (commented):
Code-wise I have two concerns (but just comments at this point):

Is the direct `bpmn-moddle` dependency really required? What do we achieve with it (in terms of features / safeguarding our users) that we don't achive by other means?

What is, in two simple lines our rules for "appliesTo" / "elementType" validation?

I'm missing a bunch of comments (examples) in the test suite / source code which makes our logic there hard to follow.

test/spec/provider/cloud-element-templates/Validator.spec.js
nikku: Well, one way to approach this would also be "allow users to do what they want" (and see how the editing experience blows up in the process) :).

But I see how we rather want to be safe. 

If we got that route though, why do we choose `bpmn:Task` and not `bpmn:Activity` as the base for all "task nodes".

marstamm:
> What is, in two simple lines our rules for "appliesTo" / "elementType" validation?

We validate that:
- The Element Template also applies to the `elementType` Target 
- The `appliesTo` and `elementType` have compatible classes (eg. no Task -> SequenceFlow allowed)

> What do we achieve with it (in terms of features / safeguarding our users) that we don't achive by other means?

We need it to check the above Assumptions. E.g., 
```
appliesTo: ['bpmn:Task']
elementType: {
  value: 'bpmn:ServiceTask'
}
```
is valid, but requires the meta-model to validate that the Service Task is a subClass of `Task`.

src/provider/element-templates/Validator.js
nikku: My proposal here is: Consider re-using the existing `moddle` instance, and its introspection capabilities: https://github.com/bpmn-io/moddle/blob/master/lib/moddle.js#L200.

nikku:
Cf. https://github.com/bpmn-io/bpmn-js-properties-panel/pull/648#discussion_r841687598.

Maybe we can re-use the existing moddle instance?

test/spec/provider/cloud-element-templates/Validator.spec.js
marstamm: > why do we choose `bpmn:Task` and not `bpmn:Activity`

Good point, I see no reason for it. I'll change it :+1:

src/provider/element-templates/Validator.js
marstamm: > Consider re-using the existing `moddle` instance

I did not consider this at first, as it was a standalone utility (no dependency injection). I will look into passing it to the constructor

src/provider/cloud-element-templates/ReplaceBehavior.js
```diff
@@ -33,7 +33,15 @@ export default class ReplaceBehavior extends CommandInterceptor {
         return;
       }
 
-      const { appliesTo } = elementTemplate;
+      const { appliesTo, elementType } = elementTemplate;
```
nikku, line 36: If this closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/642, could you link the issue from the commit?

src/provider/element-templates/Validator.js
marstamm: ...aaaand forgot to move the dependency back to dev :see_no_evil:

marstamm: > ...aaaand forgot to move the dependency back to dev see_no_evil

done with [b378bd2](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/648/commits/b378bd2fb78a9a4a0ffba95e5d27f87542de3ec6)

src/provider/element-templates/util/validate.js
```diff
@@ -5,9 +5,10 @@ import { Validator } from '../Validator';
...
  *
  * @param {Array<TemplateDescriptor>} descriptors
+ * @param {Moddle} moddle
  *
  * @return {Array<Error>}
  */
-export default function validate(descriptors) {
-  return new Validator().addAll(descriptors).getErrors();
+export default function validate(descriptors, moddle) {
+  return new Validator(moddle).addAll(descriptors).getErrors();
```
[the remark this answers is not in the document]
nikku: I thing we're fine with this.

Moddle is the _meta-model_ we rely on. If we start to validate type hierarchies (which you decided to do) there is no other way to add this.

nikku (approved):
Added https://github.com/bpmn-io/bpmn-js-properties-panel/pull/648/commits/7eb303b652a765c06c08e8e70399b2ce40a216f9 on top. Looks good to me.

PR #650, opened 2022-04-01
feat(render): display template icons in header

### LINKED ISSUE #635: Element Template Icon is shown in the Properties Panel header
__What should we do?__

The `elementTemplate.icon` should be shown in the header

image

__Why should we do it?__

Child of https://github.com/bpmn-io/internal-docs/issues/445

pinussilvestrus:
Closes #635
Depends on #641 (to work end-to-end)

### DIFF
--- a/src/render/PanelHeaderProvider.js
+++ b/src/render/PanelHeaderProvider.js
@@ -69,9 +69,7 @@ export function getConcreteType(element) {
 export const PanelHeaderProvider = {
 
   getDocumentationRef: (element) => {
-
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    const elementTemplates = useService('elementTemplates', false);
+    const elementTemplates = getTemplatesService();
 
     if (elementTemplates) {
       return getTemplateDocumentation(element, elementTemplates);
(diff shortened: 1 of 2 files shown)


src/render/PanelHeaderProvider.js
```diff
@@ -89,13 +87,21 @@ export const PanelHeaderProvider = {
   getElementIcon: (element) => {
     const concreteType = getConcreteType(element);
 
+    const elementTemplates = getTemplatesService();
+
+    if (elementTemplates) {
+      const template = getTemplate(element, elementTemplates);
+
+      if (template && template.icon) {
+        return () => <img class="bio-properties-panel-header-template-icon" width="32" height="32" src={ template.icon.contents } />;
```
marstamm, line 96: Just for my understanding: We don't enforce Sizes in the renderer, but in the Properties panel we scale everything to 32/32. If we have a non-square Icon, this will be squashed/stretched it to the 1:1 aspect ratio.

barmac (approved):

src/render/PanelHeaderProvider.js
pinussilvestrus: Yeah, just because the design in the properties panel header relies on a 32/32 icon, otherwise it would just look off. On the canvas, I assume we have more flexibility. 

We can always make it more relaxed here in the future though 👍

barmac:
Hmm I was too quick to merge this. Indeed, there is a problem with non-square icons as @marstamm pointed out: 
image

pinussilvestrus:
Damnit, you are right. I will follow up with a fixing PR!

pinussilvestrus:
The problem with `max-width` and `max-height` is that the defined template icons won't fit then, since we only have **one** **icon for canvas + header**. That's on purpose to minimize the configuration effort for template creators.

image

If we just keep the dimension, probably all template icons will look too small. Tbh, we anyway want to document best practices for template icons, e.g. that they should be square. I would therefore keep things as they are and wait for feedback on whether non-square icons are a thing.

/cc @MaxTru @nikku @currandwyer

barmac:
We could use `object-fit: contain` for proper scaling of non-square icons:

image

Original icon: 
image

It looks still OK for a square icon:

image

PR #651, opened 2022-04-04
fix(camunda-platform): add conditionalVariableName prop

marstamm:
related to https://github.com/camunda/camunda-modeler/issues/2866

image

Old Properties Panel for reference: 

image

### DIFF
--- a/src/provider/camunda-platform/properties/ConditionProps.js
+++ b/src/provider/camunda-platform/properties/ConditionProps.js
@@ -42,6 +42,12 @@ export function ConditionProps(props) {
 
   const entries = [];
 
+  if (getConditionalEventDefinition(element)) {
+    entries.push(
+      ...VariableEventProps({ element })
+    );
+  }
+
   entries.push({
     id: 'conditionType',
     component: ConditionType,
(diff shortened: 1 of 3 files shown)


src/provider/camunda-platform/properties/ConditionProps.js
```diff
@@ -333,6 +339,93 @@ function Resource(props) {
...
+      properties: {
+        variableName: value || ''
+      }
+    });
+  };
+
+  return <TextFieldEntry
+    element={ element }
+    id="conditionVariableName"
+    label={ translate('Variable Name') }
```
MaxTru, line 387: We want with sentence case with our new propPanel

```suggestion
    label={ translate('Variable name') }
```

src/provider/camunda-platform/properties/ConditionProps.js
```diff
@@ -333,6 +339,93 @@ function Resource(props) {
...
+      properties: {
+        variableEvents: value || ''
+      }
+    });
+  };
+
+  return <TextFieldEntry
+    element={ element }
+    id="conditionVariableEvents"
+    label={ translate('Variable Events') }
```
MaxTru, line 421: We want with sentence case with our new propPanel

```suggestion
    label={ translate('Variable events') }
```

src/provider/camunda-platform/properties/ConditionProps.js
```diff
@@ -333,6 +339,93 @@ function Resource(props) {
...
+function VariableEventProps(props) {
+  const entries = [];
+
+  entries.push({
+    id: 'conditionVariableName',
+    component: VariableName,
+    isEdited: isTextFieldEntryEdited
+  });
+
+  entries.push({
```
MaxTru, line 351: IMO not for StartEvents (compare existing implementation)

MaxTru (changes requested):

src/provider/camunda-platform/properties/ConditionProps.js
MaxTru: Should also add a testcase for that

marstamm: Good catch with the start event. I made it conditional

https://github.com/bpmn-io/bpmn-js-properties-panel/blob/b623f6d8c79ada663c62ace4939d658dbafc0a4e/src/provider/camunda-platform/properties/ConditionProps.js#L355-L361


And added the ConditionalStartEvent to the test case 
https://github.com/bpmn-io/bpmn-js-properties-panel/blob/b623f6d8c79ada663c62ace4939d658dbafc0a4e/test/spec/provider/camunda-platform/ConditionProps.spec.js#L537-L540

MaxTru (approved):
:wave:

PR #656, opened 2022-04-05
chore: rearrange process props by relevance

Skaiir:
Related to https://github.com/camunda/camunda-modeler/issues/2680

### DIFF
--- a/src/provider/camunda-platform/CamundaPlatformPropertiesProvider.js
+++ b/src/provider/camunda-platform/CamundaPlatformPropertiesProvider.js
@@ -44,7 +44,11 @@ import {
 const LOW_PRIORITY = 500;
 
 const CAMUNDA_PLATFORM_GROUPS = [
+  HistoryCleanupGroup,
+  TasklistGroup,
+  CandidateStarterGroup,
   ImplementationGroup,
+  ExternalTaskGroup,
   ProcessVariablesGroup,
   ErrorsGroup,
   UserAssignmentGroup,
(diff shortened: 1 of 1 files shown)


barmac:
OK so as we anticipated this moves not only process props, e.g. 
image

MaxTru:
@Skaiir , let's align with @andreasgeier and Ingo please to ensure we understand whether we want to change order GLOBALLY or ONLY FOR process

barmac:
External task is also moved: 
image

ingorichtsmeier:
Hi @barmac, 

for the timer intermediate event, your left screenshot looks as in my 5.0.0-alpha1 and it's better than the right one, as it brings the Priority under Job execution closer to the timer definition. 

Sidenote: Async continuation didn't make sense for intermediate timers as the timer are implemented as jobs. This section could be lower here, but I think only here and not for service tasks. But this is acceptable.

ingorichtsmeier:
For the external service task, your left side look different then the Modeler 5.0.0-alpha1, but also better. As the External Task appears only for Implementation type External, it should be as close as possible to Implementation.

Skaiir:
@ingorichtsmeier As close as possible but still below candidate starter, right?

ingorichtsmeier:
Yes. On the process tab, it is less important than the candidate starter.

Skaiir:
@barmac This should be a lot better.

barmac (approved):
I like it.

PR #658, opened 2022-04-05
fix(cloud-templates): move `modelerTemplateIcon` to property

marstamm:
try it out with 
```
npx @bpmn-io/sr bpmn-io/bpmn-js-properties-panel#templateIcon-as-property -c "npm run start:cloud-templates"
```

Recording 2022-04-05 at 15 18 08


marstamm:
I also checked that the deployment works, but only on `Zeebe 1.3.6` (latest in production). Is there an easy way to check for Zeebe 8?

src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
```diff
@@ -109,48 +108,13 @@ export default class ChangeElementTemplateHandler {
...
-    // create or update zeebe:modelerTemplateIcon
-    const newIcon = ensureExtension(element, 'zeebe:ModelerTemplateIcon', bpmnFactory);
-
-    return commandStack.execute('element.updateModdleProperties', {
-      element,
-      moddleElement: newIcon,
-      properties: {
-        body: contents
-      }
+    modeling.updateProperties(element, {
```
MaxTru, line 116: much simpler even :-)

MaxTru (approved):
:clap:

src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
pinussilvestrus: Yeah, how easy it is to handle attributes instead of complex extension elements 🙂

PR #659, opened 2022-04-05
fix(cloud-templates): only keep defined mappings

### LINKED ISSUE #653: Headers / inputs are not cleaned when switching between element templates
__Describe the bug__

The overall "check if properties are still valid during replace" interaction has a bug that affects changing from one element template to another where the first and second are distinct in their binding types (i.e. the first defines only inputs, the second one only headers).

In that case the properties are not cleaned up (inputs stay configured) despite the second template not having input bindings. Breaks the other way around, too.

capture ZuLYdR_optimized

__Steps to reproduce__

You can find two example templates to reproduce this [here](https://github.com/bpmn-io/bpmn-js-connectors-extension/commit/1cb6eaf8192abcf928624aa7902994e395c7caf0).

* Given I have a template that defines a task header `<a>`
* If I change to another template that does not define task headers
* Then the task header `<a>` binding is not cleaned up

I believe this is due to our C7 legacy of "only touching what has been defined"; it does not seem to be the appropriate behavior for C8 anymore.

__Expected Behavior__

Bindings (inputs, headers) are cleaned up completely, despite the new template not using them.

----

_Originally posted by @nikku in https://github.com/bpmn-io/bpmn-js-properties-panel/issues/648#issuecomment-1087471846_

marstamm:
closes #653

### DIFF
--- a/src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
+++ b/src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
@@ -255,14 +255,24 @@ export default class ChangeElementTemplateHandler {
       return newBindingType === 'zeebe:input' || newBindingType === 'zeebe:output';
     });
 
-    // (1) do not override old inputs and outputs if no new inputs and outputs specified
-    if (!newProperties.length) {
-      return;
-    }
-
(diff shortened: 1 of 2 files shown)


nikku (approved):
Great stuff :clap:

Issue #638, opened 2022-03-29
Changing element template clears existing, compatible binding values


nikku:
__Describe the Bug__

As I'm using element templates and switch between different templates (i.e. using `ElementTemplates#applyTemplate`) the implementation clears compatible values, if they match the existing template default.

![capture 1uuk08_optimized](https://user-images.githubusercontent.com/58601/160605350-b78b409b-5035-4f23-976f-41afa7ec3c70.gif)

__Steps to Reproduce__

#### template-1

```
{
  "value": "https://foobar",
  "binding": {
    "type": "zeebe:input",
    "name": "url"
  }
}
```

#### template-2

```
{
  "value": "= other",
  "binding": {
    "type": "zeebe:input",
    "name": "url"
  }
}
```

* Given I have `template-1` with `zeebe:input#url = "https://foobar` applied
* When I change to `template-2` `zeebe:input#url = "= other` is set, despite the existing binding being valid in the context of `template-2`

__Expected Behavior__

Compatible, existing binding values are kept as I change the template.

__Environment__

- Host (Browser/Node version), if applicable: Any
- OS: Any
- Library version: v1.0.0-alpha.7 (latest) but also `master`

nikku:
I'll follow up with a test case to reliably reproduce it in both C7 and C8 templates.

marstamm:
> Compatible, existing binding values are kept as I change the template.

Are we sure we want to keep existing bindings even when we did not change it from the default? I'm thinking of the scenario:

- I apply `SendGrid` Template, the default url `https://api.sendgrid.xyz/send`
- I immediately realize that I wanted a SMTP Task instead, and apply it
- URL should update to `smtp://user:host@myHost.email`

In this case, it would be weird to keep the sendgrid URL.

Implementing this would also mean that `Hidden` fields are not just hidden, but also get special behavior when applying templates. If we don't and the url input is of type `Hidden`, there is no way of knowing that the URL did not update as I would expect it until it breaks during runtime

nikku:
@marstamm Great points.

Regarding `Hidden`, you can indeed argue it is special, as it sets a fixed value (default makes no sense either).

Regarding default over existing semantics: What your comment shows is that we need to define clearly what takes precedence over what (default values over existing compatible values) before we can act here. 

---

Double checking with our existing behavior I can confirm that the original bug does not seem to be present anymore, which is great for the user experience:

[...]

PR #661, opened 2022-04-08
fix(cloud-templates): keep exisiting configuration after apply

marstamm:
Recording 2022-04-12 at 12 02 33

related to #638

### DIFF
--- a/src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
+++ b/src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
@@ -143,7 +143,7 @@ export default class ChangeElementTemplateHandler {
 
       let properties = {};
 
-      if (oldProperty && propertyChanged(changedElement, oldProperty)) {
+      if (shouldKeepValue(changedElement, oldProperty, newProperty)) {
         return;
       }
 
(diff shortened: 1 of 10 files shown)


nikku:
My proposal is that we start to test element template logic against the new APIs (`ElementTemplate#applyTemplate`) rather than internally. That will help us long term.

marstamm:
We are already testing against this API, just with some utility around it to ensure the templates are added beforehand:

https://github.com/bpmn-io/bpmn-js-properties-panel/blob/522155a8b3f849e36814805c0f78a882e15a0821/test/spec/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.spec.js#L2273-L2284

nikku:
Sorry, my bad. What I meant is "move the test to that place" (or consider not to do it).

marstamm:
I like having it as a separate file from `ElementTemplates.spec`, as we are testing the behavior of the `ChangeElementTemplateHandler` and is easier to find in the IDE this way.

But this is just a slight preference, no strong feelings either way

nikku:
> https://github.com/bpmn-io/bpmn-js-properties-panel/pull/661#issuecomment-1094700091

Great, in this case let us keep it.

I'll just make sure we have some basic coverage in the element template spec, too.

marstamm:
@nikku, how do we want to continue with this PR? From my side, the implementation for keeping configuration C8 is ready for review. Do you want to add the test cases to this PR or is this a separate issue/PR?

marstamm:
To be confirmed: this PR solved https://camunda.slack.com/archives/C02JLRNQQ05/p1649846975972899 

(double taskDefinitions after applying template to configured service task). I will add an explicit test case for this

marstamm:
Added integration tests. This is ready for review now.

nikku:
My appologies for not looking into this earlier. I have this PR on my list for tomorrow :pray:

nikku (approved):
Works as it should :+1: 

Found one unrelated visual glitch: https://github.com/bpmn-io/bpmn-js-properties-panel/issues/675.

PR #662, opened 2022-04-11
fix(templates): keep element name after remove

### LINKED ISSUE #570: Removing an element template clears its name
__Describe the Bug__

If I remove a linked element template the activities name is cleared:

capture IpqPNi_optimized

As a user I do not expect the name to be part of the element template, thus keeping it.

__Steps to Reproduce__

1. Load [element-templates.bpmn](https://github.com/bpmn-io/bpmn-js-properties-panel/files/8014624/element-templates.bpmn.txt)
2. Remove template from any element
3. See that name is removed with it

__Expected Behavior__

Name like elements that I can continue to customize as I have a template applied (group categroy value?, name, text annotation body?) are not reset as the template is being removed.

__Environment__

- Host (Browser/Node version), if applicable: Any
- OS: Any
- Library version: bpmn-js-properties-panel@1.0.0-alpha.2, `#master` (both cloud and platform providers affected)

marstamm:
closes #570

Recording 2022-04-11 at 11 28 14

### DIFF
--- a/src/provider/element-templates/util/templateUtil.js
+++ b/src/provider/element-templates/util/templateUtil.js
@@ -1,4 +1,6 @@
-import { getBusinessObject } from 'bpmn-js/lib/util/ModelUtil';
+import { getLabel } from 'bpmn-js/lib/features/label-editing/LabelUtil';
+import { createCategoryValue } from 'bpmn-js/lib/features/modeling/behavior/util/CategoryUtil';
+import { getBusinessObject, is } from 'bpmn-js/lib/util/ModelUtil';
 
 import { isUndefined } from 'min-dash';
 
@@ -21,8 +23,11 @@ export function removeTemplate(element, injector) {
   const type = businessObject.$type,
         eventDefinitionType = getEventDefinitionType(businessObject);
 
+  const newBusinessObject = createBlankBusinessObject(element, injector);
+
   const newElement = replace.replaceElement(element, {
     type: type,
+    businessObject: newBusinessObject,
     eventDefinitionType: eventDefinitionType
   });
 
@@ -88,4 +93,31 @@ function leftPad(string, length, character) {
   }
 
   return string;
+}
+
+function createBlankBusinessObject(element, injector) {
+  const bpmnFactory = injector.get('bpmnFactory'),
+        bpmnJs = injector.get('bpmnjs');
+
+  const bo = getBusinessObject(element),
+        newBo = bpmnFactory.create(bo.$type),
+        label = getLabel(element);
+
+  if (!label) {
+    return newBo;
+  }
+
+  if (is(element, 'bpmn:TextAnnotation')) {
+    newBo.text = label;
+  } else if (is(element, 'bpmn:Group')) {
+    const definitions = bpmnJs.getDefinitions();
+    const categoryValue = createCategoryValue(definitions, bpmnFactory);
+
+    newBo.categoryValueRef = categoryValue;
+    newBo.categoryValueRef.value = label;
+  } else {
+    newBo.name = label;
+  }
+
+  return newBo;
 }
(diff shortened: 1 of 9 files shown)


src/provider/element-templates/util/templateUtil.js
```diff
@@ -14,18 +15,24 @@ export function unlinkTemplate(element, injector) {
...
+        eventDefinitionType = getEventDefinitionType(businessObject),
+        label = getLabel(element);
 
   const newElement = replace.replaceElement(element, {
     type: type,
     eventDefinitionType: eventDefinitionType
   });
 
+  if (label) {
+    modeling.updateLabel(newElement, label);
```
pinussilvestrus, line 33: As we are firing a command here, the next undo after removing the template will remove the label

Kapture 2022-04-11 at 11 40 51

Only after the second undo the remove is undone. I'd expect that `undo` after the removal brings me back the template immediately.

pinussilvestrus (changes requested):
We should make sure that `undo` works properly.

nikku (commented):
Did we properly root cause this one?

With stock bpmn-js the name is kept on replace, cf.: 

capture mF96yC_optimized

How do we accomplish that (and why does it not work in connectors land)? Please root cause and we find a robust solution.

src/provider/element-templates/util/templateUtil.js
nikku: See also https://github.com/bpmn-io/bpmn-js-properties-panel/pull/662#pullrequestreview-937755048.

src/provider/element-templates/util/templateUtil.js
```diff
@@ -16,23 +16,23 @@ export function unlinkTemplate(element, injector) {
...
-        modeling = injector.get('modeling');
+        bpmnFactory = injector.get('bpmnFactory');
 
   const businessObject = getBusinessObject(element);
 
   const type = businessObject.$type,
         eventDefinitionType = getEventDefinitionType(businessObject),
         label = getLabel(element);
 
+  const newBusinessObject = bpmnFactory.create(businessObject.$type);
```
nikku, line 27: We need to be aware of different names here:

* Text annotaiton, group have different _names_

nikku: Other than that I like this direction.

marstamm: Groups are not that straight forward, but should be all good now:
https://github.com/bpmn-io/bpmn-js-properties-panel/blob/905eaf8e2b82fcd9b20925ac20b6abe30184de8d/src/provider/element-templates/util/templateUtil.js#L112-L118

src/provider/element-templates/util/templateUtil.js
```diff
@@ -89,3 +94,30 @@ function leftPad(string, length, character) {
...
+
+  const bo = getBusinessObject(element),
+        newBo = bpmnFactory.create(bo.$type),
+        label = getLabel(element);
+
+  if (!label) {
+    return newBo;
+  }
+
+  if (is(element, 'bpmn:TextAnnotation')) {
```
nikku, line 110: You could use [`LabelUtil#setLabel`](https://github.com/bpmn-io/bpmn-js/blob/develop/lib/features/label-editing/LabelUtil.js#L52) here.

nikku: And wrap it in a `try/catch` for good measure to handle categoryRef not existing, of course.

marstamm: > And wrap it in a try/catch for good measure to handle categoryRef not existing, of course

Ah, slow Github updates :smile:

marstamm: Done in [4b961e0](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/662/commits/4b961e092266ca77bb9784b3faf88818c67b44a0)

nikku (approved):
:clap:

PR #665, opened 2022-04-11
feat(templates): find all applicable templates for element

### LINKED ISSUE #604: Make it possible to get all possible template options for an element

In the old properties panel, we had a [simple util](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/0.x/lib/provider/camunda/element-templates/parts/Helper.js#L24) to get all applicable element templates for a given element. With the removal of the chooser, this API / util is not available anymore.

Furthermore, implementors have to do this on their own, e.g. 
* [the Camunda Modeler](https://github.com/camunda/camunda-modeler/blob/develop/client/src/plugins/element-templates-modal/components/ElementTemplatesModalView.js#L348)
* [the Connectors modeling demo](https://github.com/bpmn-io/connectors-modeling-demo/blob/master/connectors-extension/ConnectorsExtension.js#L139)

Provide a simple API to retrieve all element templates a given element can be applied to. Something like

```js
const elementTemplates = modeler.get('elementTemplates');

const applicableTemplates = elementTemplates.getApplicable(element);

// do whatever I want
```

/

__Additional context__

Related to https://github.com/bpmn-io/internal-docs/issues/467 (not a hard requirement)
Related to https://github.com/bpmn-io/bpmn-js-properties-panel/issues/582

marstamm:
closes #604 

This just adds a utility that can be used by upstream libraries like the element template chooser


test/spec/provider/cloud-element-templates/ElementTemplates.spec.js
```diff
@@ -139,6 +139,74 @@ describe('provider/cloud-element-templates - ElementTemplates', function() {
...
 
+  describe('getAll', function() {
+
+    it('should get all templates', inject(function(elementTemplates) {
+
+      // when
+      const templates = elementTemplates.getAll();
+
+      // then
+      expectTemplates(templates, [
```
nikku, line 150: From the API perspective a "get all (but only latest)" would be beneficial, too. 

So maybe taking inspriration from [`Overlays#get`](https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/overlays/Overlays.js#L165) with a more generic search option could work, too? Just throwing that in as a suggestion.

marstamm: I have to think about the API a bit more. I see that this would be the default use-case for Template choosers, but the only things I can come up with are
- separate functions `getAll` and `getLatest`
- additional argument to include all versions (or only latest, depending on what we decide to be the default)

nikku: Did you also check https://bpmn-io.github.io/bpmn-js-connectors-extension/?aa=1? Given the limited space for discovery (and the focus on ease of use) I'm inclined to show latest there only.

nikku: Same applies to the [element-template-chooser](https://github.com/bpmn-io/element-template-chooser) where I'm inclined to only show latest.

marstamm: At least in the demos, we don't make use of versioning.

When I think about it the other way: in what circumstances to we need older versions? I don't see a use-case outside of "old template is applied, update to latest version". Even there, we only need "what is the latest version" to compare it to the currently applied one.
Therefore, let's hide this behind a flag/additional attribute (`elementTemplates.getAll(id|element, allVersions)`).

marstamm: On second thought, it does not make sense for `getAll(id)`, which is our current use-case and would be breaking behavior. Therefore, It makes more sense to have `getLatest` alongside `getAll`, which also makes sense for `getLatest(id)`

src/provider/element-templates/ElementTemplates.js
```diff
@@ -97,13 +106,51 @@ export default class ElementTemplates {
...
+          latest: template
+        };
       }
 
       this._templates[ id ][ version ] = template;
+
+      const latestVerions = this._templates[ id ].latest.version;
+      if (isUndefined(latestVerions) || template.version > latestVerions) {
+        this._templates[ id ].latest = template;
+      }
```
marstamm, line 119: I decided to find and save the latest version during import, as we only import once but potentially check for latest versions often.

This version alias also allows us to use `get(id, 'latest')`, if we wanted to.

test/spec/provider/cloud-element-templates/ElementTemplates.spec.js
```diff
@@ -139,6 +139,157 @@ describe('provider/cloud-element-templates - ElementTemplates', function() {
...
+    it('should get latest template version (mixed versions)', inject(function(elementTemplates) {
+
+      // when
+      const templates = elementTemplates.getLatest('foo');
+
+      // then
+      expectTemplates(templates, [
+        [ 'foo', 3 ]
+      ]);
+    }));
```
marstamm, line 250: decided to go for "versioned templates are always newer than unversioned", as I see this as a possible step in the development process (first one does not have a version, second iteration does).

test/spec/provider/cloud-element-templates/ElementTemplates.spec.js
marstamm: Implemented in [fbd6d5a](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/665/commits/fbd6d5aff3f875917b36339d0223c8fd3d34113b). Ready for review :)

nikku: > Therefore, let's hide this behind a flag/additional attribute (elementTemplates.getAll(id|element, allVersions)).

Exactly. Could not have phrased that any better.

src/provider/element-templates/ElementTemplates.js
nikku: Does this work with updates (i.e. if additional templates are being loaded (async)?)

marstamm: If the templates are added using `elementTemplates.set`, then yes. This is the only API we currently offer and just allows to remove all previous and add new templates.

It does not work if it was sneakily added into `_templates` externally

src/provider/element-templates/ElementTemplatesPropertiesProvider.js
```diff
@@ -78,9 +76,7 @@ export default class ElementTemplatesPropertiesProvider {
   }
 
   _shouldShowTemplateProperties(element) {
-    return getTemplateId(element) || this._elementTemplates.getAll().some(template => {
-      return isAny(element, template.appliesTo);
-    });
+    return getTemplateId(element) || this._elementTemplates.getAll(element).length;
```
nikku, line 79: :heart:

PR #670, opened 2022-04-19
fix(render): use template version

pinussilvestrus:
Related to https://github.com/camunda/camunda-modeler/issues/2920

In the properties panel header, it now uses the template version as well to retrieve the applied template. 

**Camunda Platform 8**

image

**Camunda Platform 7**

image


src/render/PanelHeaderProvider.js
```diff
@@ -185,7 +185,8 @@ function getTemplatesService() {
 
 function getTemplate(element, elementTemplates) {
   const templateId = elementTemplates._getTemplateId(element);
-  return templateId && elementTemplates.get(templateId);
+  const templateVersion = elementTemplates._getTemplateVersion(element);
+  return templateId && elementTemplates.get(templateId, templateVersion);
```
marstamm, line 189: Why not use `elementTemplates.get(element)` instead? It supports elements as arguments

https://github.com/bpmn-io/bpmn-js-properties-panel/blob/d96930011cec7e5db04776a375fc0e4f41544878/src/provider/element-templates/ElementTemplates.js#L26-L33

pinussilvestrus: Good catch!

pinussilvestrus: I adjusted it accordingly 👍  Thanks!

src/render/PanelHeaderProvider.js
```diff
@@ -185,7 +185,7 @@ function getTemplatesService() {
 
 function getTemplate(element, elementTemplates) {
   const templateId = elementTemplates._getTemplateId(element);
```
marstamm, line 187:
```suggestion
```

src/render/PanelHeaderProvider.js
```diff
@@ -185,7 +185,7 @@ function getTemplatesService() {
...
-  return templateId && elementTemplates.get(templateId);
+  return templateId && elementTemplates.get(element);
```
marstamm, line 188:
```suggestion
  return elementTemplates.get(element);
```

marstamm (commented):
It can be simplified further: `elementTemplates.get` returns `null` when no template is set

src/render/PanelHeaderProvider.js
pinussilvestrus: That's true as we anyway check for `template && ...` anywhere we use this helper.

Issue #675, opened 2022-05-05
Default binding value is shown, where non is configured


nikku:
__Describe the Bug__

We give the user that a configured value is set in element templates when there is actually not:

![image](https://user-images.githubusercontent.com/58601/166924602-d3807ddd-fd3f-47be-a819-0ff2205740d9.png)

__Steps to Reproduce__

0. `npm run start:cloud-templates`
1. Select `groups` task
2. See that `response` is displayed in the properties panel
3. Check with BPMN 2.0 XML to see that it is not configured

__Expected Behavior__

Do not display the default value, if property is not set. Instead, render _empty_.

__Environment__

- Host (Browser/Node version), if applicable: Chrome 100
- OS: Any
- Library version: v1.0.0

PR #676, opened 2022-05-06
fix: do not render non-existing values

nikku:
Closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/675


barmac (approved):
LGTM. I think the previous behavior was just a bug.

PR #687, opened 2022-05-19
Add multiple selection state

pinussilvestrus:
Related to https://github.com/bpmn-io/properties-panel/issues/69

```sh
npx @bpmn-io/sr "bpmn-io/bpmn-js-properties-panel#use-updated-placeholder"
```

image

image

### DIFF
--- a/test/spec/BpmnPropertiesPanel.spec.js
+++ b/test/spec/BpmnPropertiesPanel.spec.js
@@ -1,4 +1,5 @@
 import {
+  act,
   render
 } from '@testing-library/preact/pure';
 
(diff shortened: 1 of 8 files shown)


test/spec/BpmnPropertiesPanel.spec.js
```diff
@@ -135,6 +177,32 @@ describe('<BpmnPropertiesPanel>', function() {
     });
 
 
+    it('should update on selection changed - multiple', async function() {
```
pinussilvestrus, line 180: Do we consider this as a breaking change (`propertiesPanel.updated` can now also contain multiple elements)? 🤔

pinussilvestrus:
Moving from https://github.com/bpmn-io/properties-panel/pull/152#issuecomment-1132682099, and concerns about the multiple message:

> We should keep our messaging as simple as possible. Interaction with elements via the canvas is obvious for users; properties panel does not need to help out with that.
> 
> What is needed for users is to understand that they need to select exactly a single element for properties editing to take place.

christian-konrad:
@pinussilvestrus 

@nikku 's point here is valid: 

We should offer the user an instruction for action how to resolve the empty state of the properties panel, therefore instead of "Use the context pad for actions" we should tell the user "Select a single element to edit its properties."

pinussilvestrus:
I bumped the properties panel version and updated the text. That should be ready for review now 👍

nikku (approved):
:+1:

PR #688, opened 2022-05-20
deps: update `properties-panel` peer-dependency

pinussilvestrus:
Missing update due to #687.


nikku (approved):

nikku:
Fell into this trap a couple of times already: https://github.com/bpmn-io/internal-docs/issues/564.

Issue #697, opened 2022-06-07
Group titles overlap element template overlays


pinussilvestrus:
__Describe the Bug__

When opening an element template overlay (e.g. "Updated available"), the group titles overlap them which breaks the design.

![image](https://user-images.githubusercontent.com/9433996/172391397-c63e07f4-7139-49cc-a52f-8a1910f6dfbc.png)

__Steps to Reproduce__

1. Execute `npx @bpmn-io/sr "bpmn-io/bpmn-js-properties-panel#master" -c "npm run start:cloud-templates"`
2. Select the "outdated" task
3. Open the "Updated available" overlay
4. See the "Custom properties" title overlaps it

__Expected Behavior__

Overlays should always be on top.

__Environment__

- Host (Browser/Node version), if applicable: Chrome 102.0.5005.61
- OS: macOS 11.6.5
- Library version: current `master`

pinussilvestrus:
It really seems like that `position: sticky` and using overlays via group header buttons do not work properly together. What we could do is disable the sticky header behavior for the element templates group only.

```css
.bio-properties-panel-templates-group .bio-properties-panel-group-header {
  position: initial;
}
```

![image](https://user-images.githubusercontent.com/9433996/172839468-f705550a-372e-41e8-98fc-702c38986ca6.png)

/cc @barmac @MaxTru @nikku 

-----

Sticky headers currently anyway don't work properly for the `ElementTemplatesGroup` as we use a custom group implementation (not using the `useStickyIntersectionObserver` hook).

![Kapture 2022-06-09 at 13 37 51](https://user-images.githubusercontent.com/9433996/172839692-9cc4c77e-0f0c-4b6b-bb4c-983fbb9a43c0.gif)

pinussilvestrus:
Moving this to ready as I think that should be fixed rather soon.

smbea:
Agreeing with Nikklas: since we have a custom implementation, the `useStickyIntersectionObserver` is not being used. And it would require some refactoring to include it because we're working a class component. For now we should just remove the sticky position for element templates - so it doesn't look broken - and tackle this later if we see fit.

PR #698, opened 2022-06-07
feat(template-util): display time for template versions

### LINKED ISSUE #694: Element Template Version with created metadata should include time
__Problem Description__

When multiple versions of a template are published at the same date, it is impossible to distinguish them and the "Update available" dropdown contains confusing information for the user as the new version date string is the same as the one of the currently used template:

image

__Solution Description__

Also include the time in the version information.

pinussilvestrus:
Closes #694

Displays the timestamp in the form _dd.MM.yyyy hh:mm_.

Bildschirmfoto 2022-06-07 um 15 07 21


barmac:
Perhaps it makes sense to display ISO time or time with time zone?

barmac:
> Working with dateTime could be anyway quite tricky. What about setting the timezone via [GitHub action](https://github.com/marketplace/actions/set-timezone)?

I'd rather set the timezone in Karma: https://blog.liprandi.net/2020/10/running-karma-in-specific-timezone.html

pinussilvestrus:
I set the timezone via karma config as @barmac suggested (https://github.com/bpmn-io/bpmn-js-properties-panel/pull/698#issuecomment-1149808012).

> Perhaps it makes sense to display ISO time with time zone?

I'd not do it, as we don't do this [in the Web Modeler](https://github.com/camunda/web-modeler/blob/main/webapp/src/App/Pages/Project/utils.js#L13) (does not set timezones there).

pinussilvestrus:
This is ready for review again.

philippfromme:
Is it just me or is the overlay not always on top?

image

pinussilvestrus:
You see correct 😆 I already created an issue for this (it's not related to this PR): https://github.com/bpmn-io/bpmn-js-properties-panel/issues/697

philippfromme (approved):
✅

PR #706, opened 2022-06-17
feat: use FEEL fields

marstamm:
requires https://github.com/bpmn-io/properties-panel/pull/158
and https://github.com/bpmn-io/feel-editor/pull/8

### DIFF
--- a/src/provider/cloud-element-templates/properties/CustomProperties.js
+++ b/src/provider/cloud-element-templates/properties/CustomProperties.js
@@ -16,7 +16,8 @@ import {
   SelectEntry, isSelectEntryEdited,
   CheckboxEntry, isCheckboxEntryEdited,
   TextAreaEntry, isTextAreaEntryEdited,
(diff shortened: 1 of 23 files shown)


marstamm:
This is ready for review. See also the [slack threat](https://camunda.slack.com/archives/GP70M0J6M/p1657706481926729) on how we want to handle changes in peer dependencies

src/provider/cloud-element-templates/properties/CustomProperties.js
```diff
@@ -138,6 +139,14 @@ function createCustomEntry(id, element, property) {
   }
 
   if (type === 'String') {
+    if (feel) {
```
barmac, line 142: Can I test it with `npm run start:cloud-templates`? If not, please add a template which allows to see it in action.

barmac:
Overall it looks good. Let's update the example and then we can merge it.

Issue #690, opened 2022-05-20
Implement Conditional Template Activation


smbea:
__What should we do?__

https://github.com/camunda/product-hub/issues/81

## Breakdown

- [x] https://github.com/camunda/element-templates-json-schema/issues/54
- [ ] Release stable version of json schema
- [x] implement for C8: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/711
  - [x] Properties panel shows entries based on condition 
  - [x] Condition mechanism  
  - [x] Adding/removing properties based on condition in xml
      - [x] on applying template 
      - [x] on value change shema
- [x] Documentation for element templates
  - [x] Update [element template docs](https://docs.camunda.io/docs/components/modeler/desktop-modeler/element-templates/defining-templates/) -https://github.com/camunda/camunda-platform-docs/pull/1008
  - [x] Update [examples]( https://github.com/camunda/camunda-modeler/blob/master/resources/element-templates/cloud-samples.json) - https://github.com/camunda/camunda-modeler/pull/3008
- [ ] To be clarified: direct documentation and guidance for Conenctors (TBD)?

__Why should we do it?__

We want to make element templates more powerful and simpler for developers.

smbea:
I spiked on the overall architecture of applying/removing conditions (with HTTP method example). This is the main points of the structure I achieved:

* **Showing entries in properties panel [`CustomProperties.createCustomEntry`]:** The entries shown in the properties panel are defined according to the properties in the template (unrelated to the xml). Therefore, we need to check that the condition is met in order to show the entry. 

``` javascript
if (isDefined(property.condition) && !checkCondition(element, template, property)) {
    // don't return entry -> entry is not applied
    return;
  }
```

* **Applying template [`ChangeElementTemplateHandler`]:** On applying template for the first time, check if condition is met for conditional properties, and if not do not apply them.

```javascript
// applies to all types of properties
_updateZeebeInputOutputParameterProperties(element, oldTemplate, newTemplate) {
...
  newProperties.forEach((newProperty) => {
    if (isDefined(newProperty.condition) && !checkCondition(element, newTemplate, newProperty)) {
       return;    // don't apply property
   }
})
... // apply property
}
```

* **Triggering condition check based on parent value change [`ChangeElementTemplateHandler.propertySetter`]**:  After updating property, check if property changed has dependant properties. If so, call propertySetter for each

```javascript
dependantProperties.forEach(property =>{
[...]

PR #711, opened 2022-06-28
Conditional templates for C8

smbea:
Related to #690 

For testing it yourself:
You can just select the "REST Method" template example (already in the default diagram) and play with the HTTP method dropdown

Here is a sketch of the architecture, which might make it easier to review:
image


src/provider/cloud-element-templates/ConditionChecker.js
```diff
@@ -0,0 +1,94 @@
...
+      element
+    } = context;
+
+    const template = this._elementTemplates.get(element);
+
+    if (!template || !this.oldTemplate) {
+      return;
+    }
+
+    const newTemplate = applyConditions(element, template);
```
nikku, line 59: Good stuff. Just like I envisioned. Does it work? :slightly_smiling_face:

smbea: It does. Just adding some more test cases and finishing touches

smbea: Ready for review

barmac:
One quirk I noticed (which is not necessarily a bug) is that sometimes we compare properties of two different templates. Let's investigate if it can cause problems. See the capture:


https://user-images.githubusercontent.com/28307541/176363156-07e2378c-97f0-417e-b3de-e43d98db9f76.mov

src/provider/cloud-element-templates/ConditionChecker.js
nikku: :smiley:

barmac:
Generally, the situation above happens each time user works with one template, and then switches to another. I wasn't able to break the mechanism with it though.

smbea:
> Generally, the situation above happens each time user works with one template, and then switches to another. I wasn't able to break the mechanism with it though.
> 
> Perhaps we should just opt-out when templates' ids are different.

Although it doesn't break anything, in this file we should only cater to changes within the same template. I added an extra condition for return

src/provider/cloud-element-templates/ConditionChecker.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  _applyConditions(context) {
+    const {
+      element
+    } = context;
+
+    const template = this._elementTemplates.get(element);
+
+    if (!template || !this.oldTemplate || template.id !== this.oldTemplate.id) {
```
nikku, line 55: We should not store the old template in `this`, but rather store it in the `context`.

Handlers shall be stateless.

src/provider/cloud-element-templates/index.js
```diff
@@ -19,10 +20,12 @@ export default {
   __init__: [
     'elementTemplatesLoader',
     'replaceBehavior',
-    'elementTemplatesPropertiesProvider'
+    'elementTemplatesPropertiesProvider',
+    'conditionChecker'
```
nikku, line 24: Can we rename this to prevent collisions, i.e. to `elementTemplatesConditionChecker`?

src/provider/cloud-element-templates/properties/CustomProperties.js
```diff
@@ -2,16 +2,15 @@ import {
...
+  isString
 } from 'min-dash';
 
-import { getBusinessObject } from 'bpmn-js/lib/util/ModelUtil';
-
 import { useService } from '../../../hooks';
 
 import { PropertyDescription } from '../../element-templates/components/PropertyDescription';
 
+import { getPropertyValue, setPropertyValue } from '../util/propertyUtil';
```
nikku, line 12: Is this cloud element template specific? If so, let's make it a dedicated cloud element templates util. Otherwise it is OK to have it in the shared util space.

test/spec/provider/cloud-element-templates/ConditionChecker.spec.js
```diff
@@ -1,9 +1,5 @@
 import TestContainer from 'mocha-test-container-support';
```
nikku, line 1: Would need to be renamed to `ElementTemplateConditionChecker.spec.js`.

test/spec/provider/cloud-element-templates/ConditionChecker.spec.js
```diff
@@ -43,13 +46,463 @@ describe('provider/cloud-element-templates - ConditionChecker', function() {
...
+        // when
+        modeling.updateProperties(element, {
+          name: 'foo'
+        });
+
+        const businessObject = getBusinessObject(element);
+        const taskDefinition = findExtension(businessObject, 'zeebe:TaskDefinition');
+
+        // then
+        expect(taskDefinition).to.not.be.undefined;
```
nikku, line 162: `to.exist`?

test/spec/provider/cloud-element-templates/ConditionChecker.spec.js
```diff
@@ -43,13 +46,463 @@ describe('provider/cloud-element-templates - ConditionChecker', function() {
...
+      })
+    );
+
+    it('should remove conditional entries', inject(
```
nikku, line 166: Add line.

test/spec/provider/cloud-element-templates/ConditionChecker.spec.js
```diff
@@ -43,13 +46,463 @@ describe('provider/cloud-element-templates - ConditionChecker', function() {
...
+      modeling.updateProperties(element, {
+        name: ''
+      });
+
+      const businessObject = getBusinessObject(element);
+
+      // when
+      commandStack.undo();
+
+      expect(findExtension(businessObject, 'zeebe:TaskDefinition')).to.not.be.undefined;
```
nikku, line 226: Replace all occurances with `to.exist`.

test/spec/provider/cloud-element-templates/ConditionChecker.spec.js
```diff
@@ -43,13 +46,463 @@ describe('provider/cloud-element-templates - ConditionChecker', function() {
...
+        modeling.updateProperties(element, {
+          name: 'bar'
+        });
+
+        const ioMapping = findExtension(businessObject, 'zeebe:IoMapping');
+        const outputs = ioMapping.get('zeebe:outputParameters');
 
-  describe('should', function() {
+        // then
+        expect(outputs).to.have.lengthOf(0);
```
nikku, line 284: `to.be.empty`.

src/provider/cloud-element-templates/properties/CustomProperties.js
smbea: It is a cloud element template specific and it's in the `cloud-element-templates/util` folder

nikku: Thanks!

nikku (approved):
Coverage issues seem to be unrelated to this PR.

Great job :clap:

PR #714, opened 2022-07-01
Add separator to process variable sources 

smbea:
Related to https://github.com/camunda/camunda-modeler/issues/3003

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2337,20 +2337,41 @@
       "dev": true
     },
     "diagram-js": {
-      "version": "8.1.1",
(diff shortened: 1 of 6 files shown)


barmac:
I can't test right now because of the error when I try to add a subprocess:

image

This seems to be caused by an incompatible combination of bpmn-js and diagram-js. Namely, we use bpmn-js@9.2.2 with diagram-js@8.1.1 even though the former library requires diagram-js@8.3.0 for [the hit boxes improvements](https://github.com/bpmn-io/bpmn-js/pull/1646).

I will push an update to the PR.

barmac (approved):
LGTM

PR #730, opened 2022-07-21
fix: add support of jQuery as parent node of panel

### LINKED ISSUE #729: jQuery Properties Panel Renderer Parent Node
__What should we do?__

Pre 1.0 one could attach the properties panel to a parent node which was provided as jQuery (see [here](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/v0.46.0/lib/PropertiesPanel.js#L411)). Now, this seems to have been deprecated (see [here](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/master/src/render/BpmnPropertiesPanelRenderer.js#L59)).
I wonder if this was a conscious decision?

__Why should we do it?__

Supporting jQuery as parent node to the properties panel would be
- practical
- in line with previous releases (see [here](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/v0.46.0/lib/PropertiesPanel.js#L411))
- especially it would go along with the way it is done in bpmn-js itself (see for example [here](https://github.com/bpmn-io/bpmn-js/blob/develop/lib/BaseViewer.js#L584)).

PaulLunkenheimer-viadee:
Closes #729

### DIFF
--- a/src/render/BpmnPropertiesPanelRenderer.js
+++ b/src/render/BpmnPropertiesPanelRenderer.js
@@ -61,6 +61,11 @@ export default class BpmnPropertiesPanelRenderer {
       throw new Error('container required');
     }
 
+    // unwrap jQuery if provided
+    if (container.get && container.constructor.prototype.jquery) {
+      container = container.get(0);
+    }
+
     if (typeof container === 'string') {
       container = domQuery(container);
     }


pinussilvestrus (approved):
Thank you ❤️

pinussilvestrus:
Merged via https://github.com/bpmn-io/bpmn-js-properties-panel/commit/bf6245a2d1fbf5d8cbc3f97767911241f516ca52 (I added the closing issue into the commit message).

PR #733, opened 2022-07-28
feat(FEEL): pass variables to FEEL entries

marstamm:
Integration of https://github.com/bpmn-io/feel-editor/issues/3 into the properties panel.

try it out with `npm start`

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
   "license": "MIT",
   "dependencies": {
     "@bpmn-io/element-templates-validator": "^0.9.0",
-    "@bpmn-io/extract-process-variables": "^0.4.5",
+    "@bpmn-io/extract-process-variables": "^0.5.0",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -63,7 +63,7 @@
     "@babel/plugin-transform-react-jsx": "^7.14.3",
     "@bpmn-io/element-template-chooser": "^0.0.5",
     "@bpmn-io/element-templates-icons-renderer": "^0.1.2",
-    "@bpmn-io/properties-panel": "^0.17.0",
+    "@bpmn-io/properties-panel": "^0.18.0",
```
pinussilvestrus, line 66: We need to update the peer dependencies as well, to avoid stuff like https://github.com/bpmn-io/internal-docs/issues/599, right?

marstamm: Good catch, I updated it 👍

Issue #728, opened 2022-07-19
[FEEL] When I copy the FEEL expression, the expression type is not copied


marstamm:
__Describe the Bug__

When I copy values from FEEL fields, the starting equal sign is missing

![Recording 2022-07-19 at 11 17 53](https://user-images.githubusercontent.com/21984219/179715163-30a728b0-cfc4-42e6-bc4e-d01594f4dab4.gif)

__Steps to Reproduce__

1. Copy Value from a feel field
2. paste it into another field
3. The FEEL type is copied

__Expected Behavior__

- FEEL is automatically enabled on the field

__Environment__

- Host (Browser/Node version), if applicable: Chrome
- OS: [e.g. Windows 7]
- Library version: dev

marstamm:
Solution Sketch:

Have a second mime type for FEEL fields `application/FEEL`, that allows us to toggle FEEL on when pasting content from another FEEL field. This way, the expected paste behavior (e.g. into notepad/emails) is not affected

marstamm:
Solved by https://github.com/bpmn-io/properties-panel/pull/174

PR #746, opened 2022-08-31
House keeping

nikku:
* Update to `npm@8`
* Incorportate `camunda-bpmn-js-behaviors@0.2.0`
* Dependency bumps

----

We should get rid of the `raw-loader` at some point, too. However the following is a common pattern in the properties-panel test suite and not supported by stock `Webpack@5` facilities:

```javascript
const diagram = require('some-path-to-diagram.bpmn').default
```

---

Closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/728


.github/dependabot.yml
```diff
@@ -14,6 +14,16 @@ updates:
     allow:
       - dependency-name: "bpmn-js"
         dependency-type: "all"
+      - dependency-name: "diagram-js"
```
pinussilvestrus, line 17: Should we add `bpmn-io/element-template-chooser` and `bpmn-io/element-template-icons-renderer` as well?

pinussilvestrus (commented):
Great stuff! Only one comment, feel free to merge afterward 👍 

Let's also follow up with https://github.com/camunda/camunda-bpmn-js/issues/158 once merged and released.

.github/dependabot.yml
nikku: Added via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/746/commits/2ced6a86ed9e216b51d7e4617b75a528220c9381.

pinussilvestrus (approved):

PR #753, opened 2022-09-06
deps: update to @bpmn-io/properties-panel@0.20.1

marstamm:
and prepare patch release for updated peer dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "@babel/plugin-transform-react-jsx": "^7.14.3",
     "@bpmn-io/element-template-chooser": "^0.0.5",
     "@bpmn-io/element-templates-icons-renderer": "^0.2.0",
-    "@bpmn-io/properties-panel": "^0.20.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -110,7 +110,7 @@
     "zeebe-bpmn-moddle": "^0.15.0"
   },
   "peerDependencies": {
-    "@bpmn-io/properties-panel": "^0.20.0",
+    "@bpmn-io/properties-panel": "^0.20.1",
```
pinussilvestrus, line 113: Do we need to be that strict, can't we simply specify `20.x`?

barmac: `^0.20.1` is in fact 0.x but not lower than 0.20.1, cf. https://docs.npmjs.com/about-semantic-versioning

marstamm: That's true, but not consistent with how we handle the others. I missed it as well

marstamm:
Closing in favour of https://github.com/bpmn-io/bpmn-js-properties-panel/pull/754, as we don't need the release for peer dependencies

PR #763, opened 2022-09-16
fix(cloud-templates): allow adding properties without default value

marstamm:
related to https://github.com/camunda/camunda-modeler/issues/3140

### DIFF
--- a/src/provider/cloud-element-templates/util/propertyUtil.js
+++ b/src/provider/cloud-element-templates/util/propertyUtil.js
@@ -5,14 +5,15 @@ import {
 } from 'min-dash';
 
 import {
+  EXTENSION_BINDING_TYPES,
   IO_BINDING_TYPES,
+  PROPERTY_TYPE,
   TASK_DEFINITION_TYPES,
   ZEEBE_TASK_DEFINITION_TYPE_TYPE,
   ZEBBE_INPUT_TYPE,
   ZEEBE_OUTPUT_TYPE,
   ZEEBE_PROPERTY_TYPE,
-  ZEEBE_TASK_HEADER_TYPE,
-  EXTENSION_BINDING_TYPES
+  ZEEBE_TASK_HEADER_TYPE
 } from '../util/bindingTypes';
 
 import {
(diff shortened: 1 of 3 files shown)


src/provider/cloud-element-templates/util/propertyUtil.js
```diff
@@ -168,8 +168,11 @@ export function setPropertyValue(bpmnFactory, commandStack, element, property, v
     property
   };
 
+  let propertyHandled = false;
```
nikku, line 171: What if we early return if handled instead? This pattern looks odd to me.

src/provider/cloud-element-templates/util/propertyUtil.js
```diff
@@ -188,6 +191,7 @@ export function setPropertyValue(bpmnFactory, commandStack, element, property, v
 
   // property
   if (type === 'property') {
```
nikku, line 193: Potential improvement: Get rid of magic value.

src/provider/cloud-element-templates/util/propertyUtil.js
marstamm: We can't return early, same as we can't switch to `if else` branches. One property might have multiple handlers within the file, which can add more commands to the `commands` array.

We could return early for a simple `property` value, as this only has one handler, but this pattern would guarantee that it works for all future properties.

nikku: Any other idea how we can re-structure this code path?

marstamm: My Proposal:
- Turn the paths into handler functions
- map property name to list of handlers
- if handler list is empty, throw error
- else: execute the commands, if any

I will try have this ready for the HoC tomorrow. Moving to "in progress" until then

src/provider/cloud-element-templates/util/propertyUtil.js
marstamm: done in [13c19df](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/763/commits/13c19df796f42a585f7204fea25678b75ae70f9c)

src/provider/cloud-element-templates/util/propertyUtil.js
marstamm: We decided to go with the NoOp solution to fix this bug, and to get rid of the `multi-command-executor` in the future, potentially during spring cleaning

Issue #764, opened 2022-09-19
Hide `deprecated` templates


nikku:
__Is your feature request related to a problem? Please describe__

As a developer I want templates to be deprecated, discouraging users from using them in the future.

__Describe the solution you'd like__

* If a template is marked as `deprecated`, don't serve it via `ElementTemplate#getLatest()` (our way to return _templates to choose from_

__Describe alternatives you've considered__

Individual components must filter templates themselves.

__Additional context__

None.

----

Child of https://github.com/bpmn-io/element-templates/issues/5.

PR #766, opened 2022-09-19
feat(element-templates): support `deprecated` templates

nikku:
Hide element templates from sight if they are marked as `deprecated`.

----

Closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/764


nikku:
Thanks for the integration test; it covered a scenario that I did not anticipate.

I fixed it and verify selection of such elements via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/766/commits/a6f44d7e4b568ac6b2eb6b7cd8dbd1bb8ed2db81.

marstamm (approved):
Works for me 🚀 

Let's get this merged, even without schema support for now. I see you already created https://github.com/camunda/element-templates-json-schema/issues/70 for this

PR #772, opened 2022-09-21
Support cron expressions for timer cycle props

barmac:
Related to https://github.com/camunda/camunda-modeler/issues/3061


philippfromme (approved):
Ideally, we'd have a shared component similar to `src/provider/shared/ExtensionPropertiesProps`. There's a lot of code duplication. But it works for now. ✔️

Issue #767, opened 2022-09-19
Replace removes valid dropdown property


nikku:
__Describe the Bug__

As I upgrade a task from one template to the other a valid property gets removed unintentionally, despite the fact it should be kept according to our good practices (`unlink + apply new template` shall always be equivalent to `replace (with new template)`).

![capture 2IJ3WT_optimized](https://user-images.githubusercontent.com/58601/190999573-3198a54c-49c1-49bc-b0cd-7dd9606a71ec.gif)

__Steps to Reproduce__

1. Go to [connectors extension demo](https://bpmn-io.github.io/bpmn-js-connectors-extension/?aa=1)
1. Open [REST Task diagram](https://github.com/bpmn-io/bpmn-js-properties-panel/files/9598153/rest-test.bpmn.1.txt)
2. Replace legacy REST Task with `REST Connector Template`
3. See that username and password entered are gone and authentication type is reset to No Auth (despite being set to `basic` in the original template)

__Expected Behavior__

* Valid template data is kept: username, password, and authentication type are kept

__Environment__

- Host (Browser/Node version), if applicable: Any
- OS: Arch
- Library version: latest (2022-09-19)

nikku:
Reproduction for this one available on [feature branch](https://github.com/bpmn-io/bpmn-js-properties-panel/compare/rest-replace-bug?expand=1).

nikku:
Removed from ready; made spring cleaning candiate.

marstamm:
## Root Cause

This behavior is actually consistent to the rules we defined:
> * New default values override old (unchanged) default values

As opposed to the unlink-apply behavior which does not know about default values and only applies:
>  * Existing values will be kept, if they are valid in the new template

Because the source field is of type "hidden", it still has the default and will be overwritten by the default of the new template.

Here are my Ideas on how to solve this:

### 1️⃣ The new Template does not define a default

This allows us to keep our rules consistent, but is not necessarily intuitive for template designers. This also requires the user to manually select a type on a new template, which is not desirable.
[Note: this is also currently broken and will replace the unchanged value with 'undefined'. I would see this as a bug]

### 2️⃣ Ditch the "default value changed" check

This solves the issue, but makes upgrade paths for all Templates more tedious for the user, as they have to check if values they never touched changed.

### 3️⃣ Add edge case for hidden values.

Basically the same we already have for `always override with hidden value`, we can have `always keep value from hidden field` when we upgrade

---
[...]

PR #796, opened 2022-09-30
fix(cloud-templates): add replace rules for dropdowns

marstamm:
closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/767

### DIFF
--- a/src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
+++ b/src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
@@ -804,8 +804,22 @@ export function findOldProperty(oldTemplate, newProperty) {
  * @returns {boolean}
  */
 function shouldKeepValue(element, oldProperty, newProperty) {
+
+  // "Hidden" values are treated as a constant
   if (newProperty.type === 'Hidden') {
     return false;
+  }
+
+  // Dropdowns should keep existing configuration
+  // cf. https://github.com/bpmn-io/bpmn-js-properties-panel/issues/767
+  if (newProperty.type === 'Dropdown') {
+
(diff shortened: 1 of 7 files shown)


src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
```diff
@@ -804,10 +804,24 @@ export function findOldProperty(oldTemplate, newProperty) {
  * @returns {boolean}
  */
 function shouldKeepValue(element, oldProperty, newProperty) {
+
+  // "Hidden" values are treated as a constant
```
nikku, line 808: If we settle down on this we should backport it to C7, right?

marstamm: In C7, we don't have these checks and use the legacy, C7 specific `propertyChanged` method. 
Adjusting it in C7 would probably be a breaking change for C7 templates and I would not see it in the scope of this PR.

https://github.com/bpmn-io/bpmn-js-properties-panel/blob/a50c9166d52cfb24fe67c8fdaec609ba4dd5e3cd/src/provider/element-templates/cmd/ChangeElementTemplateHandler.js#L1231-L1293

nikku: Perfect, thanks for clarifying. I simply forgot that we assume a more sane behavior in C8. :)

nikku:
Validated [here](https://camunda.slack.com/archives/GP70M0J6M/p1664543725921129?thread_ts=1664543274.514319&cid=GP70M0J6M).

nikku (approved):

PR #802, opened 2022-10-10
fix(cloud): remove timeCycle for interrupting timer start event

barmac:
image

This removes time cycle if start event is interrupting. Also, as part of a clean up, the code is simplified together with duration label (it used to be "Timer duration"):

image

Related to https://github.com/camunda/camunda-modeler/issues/3181

### DIFF
--- a/src/provider/zeebe/ZeebePropertiesProvider.js
+++ b/src/provider/zeebe/ZeebePropertiesProvider.js
@@ -54,7 +54,7 @@ export default class ZeebePropertiesProvider {
 
       // (2) update existing groups with zeebe specific properties
       updateMessageGroup(groups, element);
-      updateTimerGroup(groups, element);
+      updateTimerGroup(groups, element, this._injector);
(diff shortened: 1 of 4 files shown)


pinussilvestrus:
+1 for the label update. It also follows other labels, e.g. we do not say "Timer type" or "Timer value"

image

I remember we tried these shorter labels when we introduced the new properties panel. I thought something would be mentioned in [our design docs](https://github.com/camunda/modeler-design/blob/main/docs/technical-writing.md), but apparently, not.

barmac:
The CI failed because we apparently don't use `#bio-properties-panel-timerEventDefinitionValue` id for FEEL expressions.

nikku:
+1 for simplification / reducing redundancy in labels where context is clear.

nikku:
Exceptions exist (i.e. process ID vs. participant ID).

pinussilvestrus (approved):
LGTM!

PR #811, opened 2022-11-17
fix(ImplementationProps): don't remove empty properties

marstamm:
related to https://github.com/camunda/camunda-modeler/issues/3303


nikku (approved):
Very nice :star:

PR #824, opened 2022-12-01
Correct `Dropdown` rendering on undefined `zebee:input#source`

smbea:
Related to https://github.com/camunda/camunda-modeler/issues/3327, https://github.com/camunda/camunda-modeler/issues/3327#issuecomment-1333579132

### DIFF
--- a/src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
+++ b/src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
@@ -21,6 +21,7 @@ import {
   isUndefined,
   without
 } from 'min-dash';
+
 import { applyConditions } from '../Condition';
 
 /**
@@ -354,6 +355,7 @@ export default class ChangeElementTemplateHandler {
 
       // (3) add new inputs and outputs (unless optional)
       else if (shouldUpdate(newPropertyValue, newProperty)) {
+
         if (newBindingType === 'zeebe:input') {
           propertyName = 'inputParameters';
 
@@ -822,11 +824,14 @@ function shouldKeepValue(element, oldProperty, newProperty) {
     );
   }
 
+  // keep existing old property if
+  // user changed it from the original
   if (oldProperty) {
     return propertyChanged(element, oldProperty);
-  } else {
-    return !!getPropertyValue(element, newProperty);
   }
+
+  // keep existing property value
+  return !!(getPropertyValue(element, newProperty));
 }
 
 /**
--- a/test/spec/provider/cloud-element-templates/properties/CustomProperties.spec.js
+++ b/test/spec/provider/cloud-element-templates/properties/CustomProperties.spec.js
@@ -193,7 +193,7 @@ describe('provider/cloud-element-templates - CustomProperties', function() {
     }));
 
 
-    it('should change, creating zeebe:Input if non-existing', async function() {
+    it('should change, creating zeebe:TaskDefinition if non-existing', async function() {
 
       // given
       const task = await expectSelected('RestTask_noData'),
--- a/test/spec/provider/cloud-element-templates/properties/CustomProperties.spec.js
+++ b/test/spec/provider/cloud-element-templates/properties/CustomProperties.spec.js
@@ -727,6 +727,28 @@ describe('provider/cloud-element-templates - CustomProperties', function() {
       });
 
 
+      it('should display options (no visual selection)', async function() {
+
+        // when
+        await expectSelected('DropdownNoSelection');
+
+        // then
+        const entry = findEntry('custom-entry-my.example.dropdown-1-0', container),
+              options = domQueryAll('select option', entry);
+
+        expect(Array.from(options).map(({ selected, value }) => {
+          return {
+            selected,
+            value
+          };
+        })).to.eql([
+          { value: 'low', selected: false },
+          { value: 'medium', selected: false },
+          { value: 'high', selected: false }
+        ]);
+      });
+
+
       it('should change, updating binding', async function() {
 
         // given
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to [bpmn-js-properties-panel](https://github.com/bpmn-io/bpm
 
 ___Note:__ Yet to be released changes appear here._
 
+## 1.13.1
+
+* `FIX`: correctly element template defined `Dropdown` showing incorrect value
+* `DEPS`: update to `@bpmn-io/properties-panel@1.1.1`
+
 ## 1.13.0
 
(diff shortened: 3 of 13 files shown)


src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
```diff
@@ -352,8 +352,9 @@ export default class ChangeElementTemplateHandler {
         }
       }
 
-      // (3) add new inputs and outputs (unless optional)
-      else if (shouldUpdate(newPropertyValue, newProperty)) {
+      // (3) add new inputs and outputs (unless optional or undefined value)
+      else if (newPropertyValue && shouldUpdate(newPropertyValue, newProperty)) {
```
nikku, line 356: Should we test for existence of `newPropertyValue` here? It could be an empty string, too.

src/provider/cloud-element-templates/cmd/ChangeElementTemplateHandler.js
```diff
@@ -579,7 +580,7 @@ export default class ChangeElementTemplateHandler {
       }
 
       // (3) add new zeebe:Property
-      else if (shouldUpdate(newPropertyValue, newProperty)) {
+      else if (newPropertyValue && shouldUpdate(newPropertyValue, newProperty)) {
```
nikku, line 583: Should we test for existence of `newPropertyValue` here? It could be an empty string, too.

nikku (commented):
Great stuff. With this we're now consistent across the board, which is :trophy:. 

Just two minor remarks (we can ignore these and await user feedback).

nikku (changes requested):
I did a roundtrip test with the newly added `undefined-input-binding.json` template.

Result:

* Initial binding works fine (fixed in this PR)
* Removing values leads to left-overs though where the expected behavior would be to remove the bindings from XML (similar to initial creation)

Let's hold this for a moment though as I try to clarify the direction that we want to head to with Zeebe.

----

```xml
    <bpmn:serviceTask id="Activity_1m5q40l" zeebe:modelerTemplate="undefined-input-binding">
      <bpmn:extensionElements>
        <zeebe:ioMapping>
          <zeebe:input source="set-2" target="ignore" />
          <zeebe:input target="test" />
          <zeebe:output source="test" />
        </zeebe:ioMapping>
        <zeebe:taskHeaders />
        <zeebe:properties>
          <zeebe:property name="test" value="" />
        </zeebe:properties>
        <zeebe:taskDefinition />
      </bpmn:extensionElements>
    </bpmn:serviceTask>
```

nikku:
@smbea Let's proceed regarding the discovery in https://github.com/bpmn-io/bpmn-js-properties-panel/pull/824#pullrequestreview-1202578316:

> Removing values leads to left-overs though where the expected behavior would be to remove the bindings from XML (similar to initial creation).

Let's ensure on removal the whole input / property / ... is being removed, not only the value.

This way we are consistent across create + apply + removal. Properly handling input bindings in a no-magic fashion (https://github.com/camunda/zeebe/issues/11163) on the engine side is likely not built upon "empty" input.

nikku:
Related bug, found through https://github.com/bpmn-io/bpmn-js-properties-panel/pull/824#pullrequestreview-1202578316 and hopefully properly reproducible (against this branch): https://github.com/bpmn-io/bpmn-js-properties-panel/issues/830.

test/spec/provider/cloud-element-templates/properties/CustomProperties.spec.js
```diff
@@ -302,7 +302,7 @@ describe('provider/cloud-element-templates - CustomProperties', function() {
     });
 
 
-    it('should keep input (non optional)', inject(async function() {
```
nikku, line 305: Apparently by going the route I propose we violate existing contracts (non-optional inputs are always present)?

Looks like I missed on `optional` completely. Will need to assess tomorrow. :see_no_evil: 

---

For reference these are the current (web modeler) templates: https://github.com/camunda/web-modeler/tree/main/webapp/src/App/Pages/Diagram/BpmnJSExtensions/connectorsExtension/.camunda/element-templates.

nikku: > Ah yes, I thought we already had discussed this, hence "Validation indicates whether the property is needed"

We may, but I may have messed it up (missed the `optional` property), too :see_no_evil:.

Need to check what web modeler is doing, and get back to you. Sorry for that :speak_no_evil:.

nikku: Moved this to a separate PR (maybe won't fix): https://github.com/bpmn-io/bpmn-js-properties-panel/pull/833.

nikku:
As indicated [here](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/824#discussion_r1047498754) https://github.com/bpmn-io/bpmn-js-properties-panel/issues/830 is a won't fix / not a bug, as I missed `optional` completely when reproducing this issue:

* :a: Element template properties with binding `zeebe:input`, `zeebe:property` and `zeebe:output` may be marked as optional. In that case, if they are falsy (empty string) _then_ they must not be serialized into the XML. If they are not marked as optional they must be serialized into the XML as empty properties.
* :b: Optional is not supported for `zeebe:taskHeader` (why?) and `zeebe:taskDefinition:type` (makes sense?)

To validate :a: this I did the following:

* Add dedicated templates for the no variables case (with or without `optional`) :heavy_check_mark: 
* Validate it works as intended using a 
[test diagram](https://github.com/bpmn-io/bpmn-js-properties-panel/files/10227861/test.bpmn.txt)
=> Optional removes empty properties, while mandatory keeps (even if properties are broken / diagram may not deploy) :heavy_check_mark:

nikku:
Findings:

* Optional on dropdowns is not supported; I#d assume unbind (reset to empty) to be possible, when it is not (non critical): https://github.com/bpmn-io/bpmn-js-properties-panel/issues/834
* When copying conditional properties (the `Dropdown` use case this PR targets) _then_ the properties are not properly copied over (non critical; not a regression): https://github.com/bpmn-io/bpmn-js-properties-panel/issues/835
* We leave empty extension elements in the XML (non critical)

nikku:
This one is also a won't fix, as it does break the optional contract.

Actual fix for the (visual) glitch reported is https://github.com/bpmn-io/properties-panel/pull/203.

test/spec/provider/cloud-element-templates/properties/CustomProperties.spec.js
nikku: Won't fix.

nikku:
Updated this PR to contain the properties panel patch bump (https://github.com/bpmn-io/bpmn-js-properties-panel/pull/824/commits/cd1b51e60b39e6030d12569d54501053a9136b2d) + test improvements only.

nikku (approved):

PR #827, opened 2022-12-02
Implement script task as a FEEL expression

### LINKED ISSUE #825: Allow to implement Script Task as a FEEL expression
__What should we do?__

- [ ] Script task should allow to choose implementation:
  - FEEL expression
  - Job worker
- [ ]  If I select the implementation "FEEL expression" then I need to define the FEEL expression to evaluate and a variable name to store the result in. Both fields are required. It should work similarly to a business rule task.
- [ ]  The script task should define either an inline FEEL expression or a job type. But not both.

__Why should we do it?__

Part of https://github.com/camunda/camunda-modeler/issues/3321

barmac:
Closes #825

### DIFF
--- a/src/provider/zeebe/ZeebePropertiesProvider.js
+++ b/src/provider/zeebe/ZeebePropertiesProvider.js
@@ -12,6 +12,8 @@ import {
   MultiInstanceProps,
   OutputPropagationProps,
   OutputProps,
+  ScriptImplementationProps,
+  ScriptProps,
   TargetProps,
   TaskDefinitionProps,
   TimerProps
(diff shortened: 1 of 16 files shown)


barmac:
Hmm I've just noticed that "isEdited" does not work correctly in the implementation group when FEEL expression is selected 
image

barmac:
The last bug is fixed via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/827/commits/af72b2c9b4dead3a79bab86dc301b3961cfe8d12

This PR is ready to review.

nikku:
Love last bugs fixed :heart:

Issue #838, opened 2022-12-15
Input mappings defined by templates not stable


nikku:
__Describe the Bug__

As a user I define element template properties in a dedicated order. As it looks like that order is not uphelt as input mappings are created. In fact it seems like we attach the last saved last in the `<zeebe:ioMapping>` list. This leads to dependent mappings not being evaluated consistently (https://github.com/camunda/web-modeler/issues/3251). 

![capture NxgZqR_optimized](https://user-images.githubusercontent.com/58601/207800985-da95607c-1a2b-4308-934a-e60b250ca815.gif)

The whole behavior is bogus to the user, and cannot worked around without knowing.

__Steps to Reproduce__

1. configure [`dependent-io.json` template](https://github.com/bpmn-io/bpmn-js-properties-panel/files/10234754/dependent-io.json.txt)
2. Model `Dependent Mapping` task
2. Change values of `First*`
3. See that we move each one of them (`zebee:input`, `zeebe:output`, `zeebe:property`) below dependent property

__Expected Behavior__

* Upon modification we uphold the order of properties defined in the template.
* Upon entry addition (i.e. conditionally enabled) we uphold the order of properties defined in template.

__Environment__

- Host (Browser/Node version), if applicable: Any
- OS: Any
- Library version: 1.12.0

nikku:
I've investigated the issue and concluded this is not an easy fix. 

Added [failing test cases](https://github.com/bpmn-io/bpmn-js-properties-panel/compare/838-stable-property-order), and moving to `ready` to be picked up in an orderly manner next year.

nikku:
Adding the blocker label just now. 

This impacts Camunda connectors in a large scale, as it prevents real-world re-use of inbound webhook and outbound REST/JSON ([ref](https://camunda.slack.com/archives/C0350959WG1/p1671453655773349), [ref](https://camunda.slack.com/archives/C0350959WG1/p1671484548468209)). We should pick this up with priority in early 2023.

CC @CatalinaMoisuc @christian-konrad

MaxTru:
> Adding the blocker label just now.
> 
> This impacts Camunda connectors in a large scale, as it prevents real-world re-use of inbound webhook and outbound REST/JSON ([ref](https://camunda.slack.com/archives/C0350959WG1/p1671453655773349), [ref](https://camunda.slack.com/archives/C0350959WG1/p1671484548468209)). We should pick this up with priority in early 2023.
> 
[...]

Issue #506, opened 2021-11-29
Auto-scroll to newly created list items


mschoe:
__Is your feature request related to a problem? Please describe__

New list items are not visible when I create them at the lower part of the protperties panel. In this case I have to scroll  manually down to the see the full item.

https://user-images.githubusercontent.com/2777965/142612470-510427e2-2947-4bc6-9b84-b8bc89e3a66d.mp4

__Describe the solution you'd like__

It would be nice if the properties panel automatically focuses the new list item and manually scrolling is not necessary. 

__Describe alternatives you've considered__

there is no alternative ;)

__Additional context__

PR #843, opened 2023-01-03
feat: don't sort input output mappings visually

MaxTru:
Currently ioMappings are visually sorted.

This is a problem, because the sorting in the XML has a semantic. See https://docs.camunda.io/docs/components/concepts/variables/#inputoutput-variable-mappings: 

> Variable mappings are evaluated in the defined order. Therefore, a source expression can access the target variable of a previous mapping.

This is highly relevant in Connectors land: we need to define certain variables first, and reference them then later.

Hence I propose this as a quick fix: with this PR the sorting is at least visible in the UI. As potential follow-up, we could support a reordering via the prop panel.

related to #838

### DIFF
--- a/src/provider/zeebe/properties/InputProps.js
+++ b/src/provider/zeebe/properties/InputProps.js
@@ -48,7 +48,8 @@ export function InputProps({ element, injector }) {
 
   return {
     items,
-    add: addFactory({ element, bpmnFactory, commandStack })
+    add: addFactory({ element, bpmnFactory, commandStack }),
+    shouldSort: false
   };
 }
 
--- a/src/provider/zeebe/properties/OutputProps.js
+++ b/src/provider/zeebe/properties/OutputProps.js
@@ -48,7 +48,8 @@ export function OutputProps({ element, injector }) {
 
   return {
     items,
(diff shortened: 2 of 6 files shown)


marstamm (approved):
Thank you for the PR! Makes sense to not order them by label. I tested it and it works fine.

The default IO-mappings are also not affected by https://github.com/bpmn-io/bpmn-js-properties-panel/issues/838, so we can merge it right away

MaxTru:
@christian-konrad are there any concerns from PM side?

nikku:
I don't regard this a feature, but a bug fix and you provided a great rationale for that.

If we want to improve (solve the user problem of "specific ordering") we should not order automatically, but allow users to re-order manually, as you @MaxTru suggested.

nikku:
Just a heads-up that https://github.com/bpmn-io/bpmn-js-properties-panel/issues/506 is becoming a more pressing issue now (again):

capture eDXlp2_optimized

Overall I love this interaction though. It is much more natural ("as I would expect").

nikku:
Backported to C7 via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/845.

PR #847, opened 2023-01-05
test: correct test container styles

nikku:
Ensure test container is full size, without hacks :tm:

The key is to make the test content container "as high as parent" using `overflow: hidden`. All content will follow suite naturally once that is settled.

capture mONZvA_optimized

### DIFF
--- a/test/test.css
+++ b/test/test.css
@@ -1,36 +1,35 @@
 @import url('https://fonts.googleapis.com/css2?family=IBM+Plex+Sans:wght@300;400;500;600&display=swap');
 
-html, body, .modeler-container, .modeler-container > div {
+* {
+  box-sizing: border-box;
+}
+
+html,
+body {
   height: 100%;
   margin: 0;
   font-family: 'IBM Plex Sans', sans-serif;
 }
 
 .test-container {
-  height: 800px !important;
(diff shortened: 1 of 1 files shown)


test/test.css
```diff
@@ -1,36 +1,35 @@
...
+  height: 100vh !important;
 }
 
 .test-content-container {
-  width: 100%;
-  height: calc(100% - 24px) !important;
   display: flex;
   flex: 1;
   flex-direction: row;
+  overflow: hidden;
```
marstamm, line 24: Is there a specific case where `overflow: hidden` is needed? I tried it out without it and it still works fine without it for me.

nikku: It is needed to not automatically expand, i.e. with properties panel expanding.

nikku: As I mentioned in the PR: 

> The key is to make the test content container "as high as parent" using overflow: hidden. All content will follow suite naturally once that is settled.

capture X8cWwS_optimized

marstamm: Oh, okay. It get's longer when the properties panel expands, I missed that. Thank you for elaborating

marstamm (approved):

PR #855, opened 2023-01-18
feat: allow escalation code to be a FEEL expression

barmac:
Related to https://github.com/camunda/camunda-modeler/issues/3318


barmac:
I should update this according to https://github.com/bpmn-io/bpmn-js-properties-panel/pull/856

barmac:
> I should update this according to #856

Never mind, it's already there.

barmac:
@marstamm @smbea do you need intro for the review?

PR #864, opened 2023-02-02
fix: make `name` input multi-line

### LINKED ISSUE #705: `Name` field should ne multi-line
__Describe the Bug__

Name Fields used to be TextAreas in version 0.x of the properties panel. Multi-line Names can not be correctly displayed in the properties panel this way

image

__Steps to Reproduce__

1. Create a Task
2. Use direct editing to give it a name with line break
3. The properties panel does not show the line breaks

__Expected Behavior__

Line breaks are shown correctly

__Environment__

- Host Chrome 100
- OS: Linux
- Library version: dev

smbea:
Closes #705

### DIFF
--- a/src/provider/bpmn/properties/NameProps.js
+++ b/src/provider/bpmn/properties/NameProps.js
@@ -9,7 +9,7 @@ import {
   add as collectionAdd
 } from 'diagram-js/lib/util/Collections';
 
-import { TextFieldEntry, isTextFieldEntryEdited } from '@bpmn-io/properties-panel';
+import { TextAreaEntry, isTextAreaEntryEdited } from '@bpmn-io/properties-panel';
 
 import {
   useService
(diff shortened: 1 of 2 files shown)


nikku (commented):
Looks fine code wise.

UX wise I'm wondering if it is time to make [textareas auto-resizable](https://stackoverflow.com/a/24676492) (to a certain point):

capture KBNS88_optimized

src/provider/bpmn/properties/NameProps.js
```diff
@@ -64,7 +64,8 @@ function Name(props) {
     },
     getValue: (element) => {
       return element.businessObject.name;
-    }
+    },
+    rows: 1
```
nikku, line 68: I'd rather have an auto-resizing mechanism in place here (up to a certain height).

nikku (approved):

PR #867, opened 2023-02-02
Don't sort headers/extension properties

### LINKED ISSUE #865: Extension properties / headers should not sort automagically
### Is your feature request related to a problem? Please describe.

Related to https://github.com/bpmn-io/bpmn-js-properties-panel/pull/843 we should not sort technical properties automagically:

capture ACtmBs_optimized

### Describe the solution you'd like

Technical properties (headers, extension properties) are always displayed in insertation order. This way we limit surprises for our users. Especially with extension properties we have no guarantees how users use them (and if order matters).

### Describe alternatives you've considered

None.

### Additional context

None.

smbea:
Closes #865

### DIFF
--- a/src/provider/zeebe/properties/HeaderProps.js
+++ b/src/provider/zeebe/properties/HeaderProps.js
@@ -46,7 +46,8 @@ export function HeaderProps({ element, injector }) {
 
   return {
     items,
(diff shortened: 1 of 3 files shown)


nikku (approved):
Amazing :tada:

Issue #840, opened 2022-12-20
Support `optional` for `zeebe:taskHeader` bindings


nikku:
__Is your feature request related to a problem? Please describe__

When creating C8 templates `zeebe:taskHeader` cannot be marked as `optional`. As a result users cannot use the `optional` pattern consistently to "only serialize what is user provided" (cf. https://github.com/camunda/team-connectors/issues/282):

![capture pxSok2_optimized](https://user-images.githubusercontent.com/58601/208639592-ee5c29e2-d86c-4880-84d4-45f229324283.gif)

__Describe the solution you'd like__

* `optional` is supported for `zeebe:taskHeader`

__Describe alternatives you've considered__

None.

__Additional context__

This is a UX issue; once I know `optional` it is not understandable to me as a template author why it should not be supported for task headers.

nikku:
CC @christian-konrad

PR #868, opened 2023-02-03
feat: Support `optional` for `zeebe:taskHeader` bindings

marstamm:
closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/840

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.16.0",
       "license": "MIT",
       "dependencies": {
-        "@bpmn-io/element-templates-validator": "^0.11.0",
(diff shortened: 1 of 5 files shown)


nikku (changes requested):
Do we already implement the `optional` logic here?

marstamm:
> Do we already implement the `optional` logic here?

Optional is already implemented in the property update logic. This change only required a schema update 
https://github.com/bpmn-io/bpmn-js-properties-panel/blob/9f798371f863b466486242d9ff224f9bf59badc7/src/provider/cloud-element-templates/util/propertyUtil.js#L350-L351

Issue #858, opened 2023-01-27
Improve representation of non FEEL languages


nikku:
### Is your feature request related to a problem? Please describe.

As more people adapt element templates for domain-specific configuration we'll see the need to embed non-FEEL going to see more extensions for non FEEL domain specific languages. While I don't necessarily expect full editor tooling I do expect monospace editing and decent sizing of the input.

**Example:** Embed a [GraphQL query / mutation editor](https://github.com/camunda/connectors-bundle/pull/202) in an element template:

![image](https://user-images.githubusercontent.com/58601/215203643-4bd28b32-41b4-4913-bc24-ec73f09df8a4.png)

### Describe the solution you'd like

* [x] Support text editors other than FEEL
  * [x] Display content as monospace font (https://github.com/bpmn-io/bpmn-js-properties-panel/pull/869)
  * [x] Ensure editor grows with content (https://github.com/bpmn-io/bpmn-js-properties-panel/pull/870)
  * [ ] #871

### Describe alternatives you've considered

Leave things as is.

### Additional context

Brought up in the context of https://github.com/camunda/connectors-bundle/pull/202.

nikku:
Idea: 

* Allow `lang` or `type` to be define on `textarea` input bindings
  ```json
  {
    "name": "My special Template
    ...
    "properties: [
      {
        "label": "Query/Mutation",
        "type": "Text",
        "language": "graphql",
        "binding": {
          "type": "zeebe:input",
          "name": "graphql.query"
        }
      }
    ],
    ...
  }
  ```
* Allow to plug in editors (with a well defined contract) for a `type`
* Default to a (reasonable) monospace editor fallback if no editor for defined `type` is found.

marstamm:
My Thoughts on this:
- From a user perspective, setting the language would be the best, I agree.
- While the plugin system sounds good, I'm not sure if users will actually use it or if we could just have a lightweight multi-editor in the properties-panel. syntax highlighting for common languages could either be achieved by reusing codemirror or with a light-weight regex based solution

I think we can already move ahead and improve the user experience without a full picture around the plugin API:
- Allow the `language` property in the JSON schema
- when it is used, set `monospace` on the default Text area https://github.com/bpmn-io/properties-panel/blob/main/src/components/entries/TextArea.js#L26
- Adjust TextArea behavior to auto-expand

nikku:
> I think we can already move ahead and improve the user experience without a full picture around the plugin API:

[...]

PR #869, opened 2023-02-03
feat: render `Text#language` fields using monospace font

nikku:
screenshot XsLXf7

Related to https://github.com/bpmn-io/bpmn-js-properties-panel/issues/858


marstamm (approved):
Great improvement! I take it the [schema support](https://github.com/camunda/element-templates-json-schema/pull/89) will be added through a dependency update later?

nikku:
> I take it the https://github.com/camunda/element-templates-json-schema/pull/89 will be added through a dependency update later?

YES. It is not strictly necessary.

PR #875, opened 2023-02-08
No sort C7

nikku:
This ensures that remaining C7 items are not sorted where this is unexpected.


nikku:
Closed in favor of https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1047.

PR #877, opened 2023-02-09
Apply default template on initial creation only

nikku:
Before this PR we did not properly account for the `createElementBehavior=false` hint that is provided whenever shape creation should be side-effect free ("as is"), i.e. during paste. See also https://github.com/camunda/camunda-modeler/issues/3235#issuecomment-1424188942.

This PR fixes that behavior, and also addresses https://github.com/camunda/camunda-modeler/issues/3235 in the process.


philippfromme (approved):
✅

PR #878, opened 2023-02-10
fix(cloud-element-templates): `createElement` conditional props

smbea:
Related to https://github.com/bpmn-io/internal-docs/issues/718

The explanation of this approach is [here](https://github.com/bpmn-io/internal-docs/issues/718#issuecomment-1423938684)

### DIFF
--- a/src/provider/cloud-element-templates/Condition.js
+++ b/src/provider/cloud-element-templates/Condition.js
@@ -16,7 +16,7 @@ export function applyConditions(element, elementTemplate) {
   };
 }
 
-function isConditionMet(element, properties, property) {
+export function isConditionMet(element, properties, property) {
   const { condition } = property;
 
   // If no condition is defined, return true.
@@ -43,7 +43,7 @@ function isSimpleConditionMet(element, properties, condition) {
   return false;
 }
 
-function getValue(element, properties, propertyId) {
+export function getValue(element, properties, propertyId) {
   const property = properties.find(p => p.id === propertyId);
 
   if (!property) {
--- a/src/provider/cloud-element-templates/create/TemplateElementFactory.js
+++ b/src/provider/cloud-element-templates/create/TemplateElementFactory.js
@@ -23,7 +23,9 @@ import {
   ZEBBE_PROPERTY_TYPE
 } from '../util/bindingTypes';
 
-import { applyConditions } from '../Condition';
+import {
+  isConditionMet
+} from '../Condition';
 
 export default class TemplateElementFactory {
 
@@ -52,13 +54,12 @@ export default class TemplateElementFactory {
 
     const {
       appliesTo,
-      elementType
+      elementType,
+      properties
     } = template;
 
     const elementFactory = this._elementFactory;
-    const bpmnFactory = this._bpmnFactory;
     const moddle = this._moddle;
-    const providers = this._providers;
 
     // (0) make sure template is valid
     const errors = validate([ template ], moddle);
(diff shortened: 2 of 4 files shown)


src/provider/cloud-element-templates/Condition.js
```diff
@@ -30,7 +30,17 @@ function isConditionMet(element, properties, property) {
...
+  let propertyValue;
+
+  if (element) {
+
+    // applying template to element
+    propertyValue = getValue(element, properties, property);
+  } else {
+
+    // creating element with template: use default value
+    propertyValue = getDefaultValue(properties, property);
```
nikku, line 42: The one issue that I see with this approach is that we account for properties (and their value) that are conditionally deactivated (won't otherwise be applied). We'd need to solve that "conflict" one way or the other.

A sane alternative to what you implemented would be:

* Define a clear processing order => Properties are established first to last
* For each property in definition order check if it is valid against the currently established binding values => If yes, establish binding, if not ignore.

The benefit of that approach is that we could move the logic out of this utility, to the [template factory](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/878/files#diff-3e86785fc01cbc4b7bd468798cec81083d46dfcffa00f780d25500b850b0c4c7). This utility could remain as is. It simply checks the condition against currently bound values available in the element.

nikku: You sketch a third approach in https://github.com/bpmn-io/internal-docs/issues/718#issuecomment-1423938684: Mirror the existing update behavior.

(1) first apply all properties
(2) then hide all properties that no longer apply (per configuration, executed from top to bottom (first to last property))

nikku (changes requested):
https://github.com/bpmn-io/bpmn-js-properties-panel/pull/878/files#r1103308694

src/provider/cloud-element-templates/Condition.js
nikku: > To achieve the existing behaviour means changing the [create helpers](https://github.com/bpmn-io/bpmn-js-properties-panel/tree/master/src/provider/cloud-element-templates/create) to use the modeling.updateProperties API

Not necessarily. We would not _re-use_ but simply ensure that what is being executed (and how filtering happens) works in the exact same way:

* First we setup the element (one step, all properties)
* Then we remove properties that are not valid

nikku: My bad, I did not get that right about pre-filtering.

> What you are describing would mean implementing this part (bellow) again?

I'm not suggesting any implementation. I'm just suggesting that we align both behaviors and have _one_ defined behavior. If that is already the case / in place, that is great. 

If it is not in place yet we want to get there, either sharing the algorithm _or_ at implementing it in the same spirit. As creating an element from a template is not invoking the `CommandStack` sharing won't be easily possible, I think.

nikku: With https://github.com/bpmn-io/bpmn-js-properties-panel/pull/878#discussion_r1104591164 I was referring to the update behavior, where we first set, and then remove properties.

Seems conditional behavior is a little bit all over the place :).

nikku: We further discussed this issue and found that the algorithm for safe creation of properties boils down to (1) create a dependency tree (2) establish properties in order of that tree, going from no dependencies -> dependent on most things:

property-binding excalidraw

nikku:
Great that you followed up on this :star2:.

I really like the approach and added https://github.com/bpmn-io/bpmn-js-properties-panel/pull/878/commits/eaeca6164088ace29dbe4de3b59051343f8c5331 as a (hopefully) sensible improvement on top. Some basic improvements making the code easier to maintain / reason about I hope. 

Please have a look and tell me what you think.

nikku (approved):

PR #891, opened 2023-03-16
feat: persist layout state in localStorage

marstamm:
Recording 2023-03-16 at 11 43 35

### DIFF
--- a/src/render/BpmnPropertiesPanel.js
+++ b/src/render/BpmnPropertiesPanel.js
@@ -1,7 +1,8 @@
 import {
   useState,
   useMemo,
-  useEffect
+  useEffect,
+  useCallback
 } from '@bpmn-io/properties-panel/preact/hooks';
 
 import {
@@ -32,7 +33,7 @@ export default function BpmnPropertiesPanel(props) {
     element,
     injector,
     getProviders,
-    layoutConfig,
+    layoutConfig: initialLayoutConfig,
     descriptionConfig
   } = props;
 
@@ -188,11 +189,24 @@ export default function BpmnPropertiesPanel(props) {
   }, [ providers, selectedElement ]);
 
   // (5) notify layout changes
-  const onLayoutChanged = (layout) => {
+  const [ layoutConfig, setLayoutConfig ] = useState(initialLayoutConfig || {});
+
+  const onLayoutChanged = useCallback((newLayout) => {
     eventBus.fire('propertiesPanel.layoutChanged', {
-      layout
+      layout: newLayout
     });
-  };
+  }, [ eventBus, layoutConfig ]);
+
+  // React to external layout changes
+  useEffect(() => {
+    const cb = (e) => {
+      const { layout } = e;
+      setLayoutConfig(layout);
+    };
+
+    eventBus.on('propertiesPanel.setLayout', cb);
+    return () => eventBus.off('propertiesPanel.setLayout', cb);
+  }, [ eventBus, setLayoutConfig ]);
 
   // (6) notify description changes
   const onDescriptionLoaded = (description) => {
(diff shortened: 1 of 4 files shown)


test/spec/BpmnPropertiesPanel.spec.js
```diff
@@ -314,7 +401,7 @@ describe('<BpmnPropertiesPanel>', function() {
       createBpmnPropertiesPanel({ container, eventBus });
 
       // then
-      expect(loadedSpy).to.have.been.calledOnce;
+      expect(loadedSpy).to.have.been.called;
```
marstamm, line 404: This basically tested internal API of the properties-panel before, cf. https://github.com/bpmn-io/properties-panel/pull/228

src/hooks/useSavedLayout.js
```diff
@@ -0,0 +1,59 @@
...
+    return () => eventBus.off('propertiesPanel.layoutChanged', cb);
+  }, [ eventBus ]);
+
+
+  return layout;
+}
+
+// helpers //////
+function getSavedLayout() {
+  const savedLayout = window.localStorage.getItem('bpmn-properties-panel-layout') || '{}';
```
nikku, line 48: I'm not sure if we want to hard code this inside the properties panel. 

I'd rather

* Provide a clear api / contract to listen for properties panel layout changes
* Instantiate the properties panel or update it with a shared layout

:arrow_right: Be API first, and not bury this inside the properties panel. How saving happens (if remotely or locally) is an application concern I think.

nikku: If we provide such a hook then could we make saving and getting configurable to accomplish https://github.com/bpmn-io/bpmn-js-properties-panel/pull/891/files#r1140288946?

marstamm: I agree that the API should be explicit. I added [a172bd0](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/891/commits/a172bd038031babb9d6fed7f35f7c73d6fdbb70b) to make the API more explicit and also expose it as `propertiesPanel.setLayout()`.

I disagree that the implementation should always be on the integration side. In my opinion, it makes sense to have sensible defaults in the core which can be overwritten by the application. This is the case here: you can implement your own persistence/update logic by subscribing to `propertiesPanel.layoutChanged` and using the API (either eventBus or `propertiesPanel.setLayout`). 
If you don't implement your own handler, you still get basic persistence across tabs in localStorage as the default implementation.

nikku: Let's say that this would be the first time that we move persistent state inside our libraries :).

marstamm: Fair point. Removed with [c82d5b3](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/891/commits/c82d5b376084c99a3a7d461c7a7458aaa21e1f1e)

marstamm:
My last commit on Friday broke the tests, it is fixed now :)

Ready for review again

src/render/BpmnPropertiesPanel.js
```diff
@@ -207,7 +221,7 @@ export default function BpmnPropertiesPanel(props) {
       headerProvider={ PanelHeaderProvider }
       placeholderProvider={ PanelPlaceholderProvider(translate) }
       groups={ groups }
-      layoutConfig={ layoutConfig }
+      layoutConfig={ layout }
```
nikku, line 224: Would make sense to change this key to `layout`, too? 

Otherwise let's stay consistent and keep `layout` named `layoutConfig`.

nikku (approved):
Works for me.

marstamm:
actually, let's move this to WIP until https://github.com/bpmn-io/properties-panel/pull/228 is merged and released. We intermittently removed the feature this PR relies on with `@bpmn-io/properties-panel@1.5.1`

PR #902, opened 2023-03-29
Only variable complete fields backed by IO mappings

nikku:
This prevents us from providing bogus suggestions.

Related to https://github.com/camunda/camunda-modeler/issues/3525

----

capture 8TR03a_optimized


nikku:
@marstamm Do you have a specific way in mind how I can test this? I'd otherwise just leave it as manually tested for the time being.

marstamm:
Looking into it right now

marstamm:
I found no easy way to test this without stubing/dynamically mocking modules, for which we currently don't have the infrastructure.

I made some changes so that the Editor still uses the TooltipContainer provided. Other than that, I'm okay with merging it as-is

nikku:
Great. Thanks for having a look.

nikku:
I'll go ahead and release this.

Issue #890, opened 2023-03-15
Support intermediate message catch event templates


barmac:
### What should we do?

In order to accomplish https://github.com/camunda/camunda-modeler/issues/3403, we need to implement proper support in the element templates infrastructure:

* [x] Create event with message event definition from template
* [x] Replace already existing event definition when template is applied
* [x] Support `bpmn:Message` bindings:
  * [x] property
  * [x] zeebe:subscription
  * [x] as unconditional properties
  * [x] with conditions
* [x] Generate unique message name via `generatedValue: 'uuid'`
  * [x] apply template
  * [x] create from template
* [x] Hide templated message from messages list
* [x] When template is removed, remove message altogether
* [x] Apply should create a new message instead of reusing the referenced one
* [x] Copy and paste in the same diagram should create a new message
* [x] Replace with standard should remove the message
* [x] Handle template change (old and new template specified)
* [x] Remove message when element is removed
* [x] Message select should stay on top of the group
* [x] (optional) https://github.com/bpmn-io/bpmn-js/issues/1872

### Why should we do it?

Related to https://github.com/camunda/camunda-modeler/issues/3403

barmac:
In the current state, we don't set $parent on message. I will look into this in spare time.

barmac:
I will implement a fix for https://github.com/bpmn-io/bpmn-js/issues/1872 as this is an annoying blocker for the conditional properties.

barmac:
Follow-up:
* https://github.com/bpmn-io/bpmn-js-properties-panel/issues/914
* https://github.com/bpmn-io/bpmn-js-properties-panel/issues/915

I did not create an issue for magic string as it's a refactoring.

PR #906, opened 2023-04-13
Support message event templates

barmac:


barmac:
This is now rebased on `master`.

barmac:
I just noticed that we don't update `modelerTemplate` on Message when template is replaced.

barmac:
Issues/potential topics that @philippfromme and I (re)discovered in today's session:
1. Generated value should be regenerated on paste
2. Message should be removed if no message properties exist on template
3. MessageProps should use constant instead of magic string `zeebe:modelerTemplate`
4. Unlink should not remove message from the diagram but only remove the `zeebe:modelerTemplate` property.

philippfromme (approved):
Let's revisit the follow-ups soon and iterate. ✅

barmac:
I'll create issues

barmac:
Done: https://github.com/bpmn-io/bpmn-js-properties-panel/issues/890#issuecomment-1527278756

PR #919, opened 2023-05-09
fix: unlink templated message instead of removing

### LINKED ISSUE #914: Unlink should not remove message from the diagram but only remove the `zeebe:modelerTemplate` property
### What should we do?

In cloud-element-templates, we remove the message completely if an event is unlinked. We should unlink the message instead.

### Why should we do it?

To follow the unlink pattern.

barmac:
Closes #914


marstamm (approved):
Makes sense

PR #922, opened 2023-05-15
feat: validate `dropdown` and `textarea` custom entries

marstamm:
related to https://github.com/camunda/camunda-modeler/issues/3567

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "@babel/plugin-transform-react-jsx": "^7.14.3",
     "@bpmn-io/element-template-chooser": "^1.0.0",
     "@bpmn-io/element-templates-icons-renderer": "^0.3.0",
-    "@bpmn-io/properties-panel": "^1.3.1",
+    "@bpmn-io/properties-panel": "^1.8.2",
     "@bpmn-io/variable-resolver": "1.0.1",
     "@rollup/plugin-alias": "^5.0.0",
     "@rollup/plugin-babel": "^6.0.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -64,7 +64,7 @@
     "@babel/plugin-transform-react-jsx": "^7.14.3",
     "@bpmn-io/element-template-chooser": "^1.0.0",
     "@bpmn-io/element-templates-icons-renderer": "^0.3.0",
-    "@bpmn-io/properties-panel": "^1.3.1",
+    "@bpmn-io/properties-panel": "^1.8.2",
```
nikku, line 67: :see_no_evil:

test/spec/provider/cloud-element-templates/properties/CustomProperties.json
```diff
@@ -255,10 +274,197 @@
...
+          "name": "name"
+        },
+        "constraints": {
+          "pattern": {
+            "message": "Must be positive integer",
+            "value": "\\d+"
+          }
+        }
+      },
+      {
```
nikku, line 373: I don't see where we validate this. Could you help me out?

nikku: Ah, it is https://github.com/bpmn-io/bpmn-js-properties-panel/pull/922/files#diff-b45460173a6452b61501a862ae9797960a113256f389211874da60d2f98ee020R1239.

nikku (approved):
Sounds good. 

Pattern validation for dropdown seems to me a little unexpected. But then again we could argue that `/FOO/` as a validator pattern + not being available as a choice is a template error.

barmac:
This is actually breaking the properties panel for certain cases, cf. https://camunda.slack.com/archives/CKGH9LR40/p1684417089472889
Basically the validator receives sometimes `undefined` and it fails thereafter.

PR #930, opened 2023-05-23
Connectors extension

nikku:
Uses `bpmn-js-create-append-anything` in favor of the now phased out connectors extension.


smbea (approved):
nice

PR #935, opened 2023-06-08
Add utilities to api

smbea:
This PR adds the `removeTemplate`, `unlinkTemplate` and `updateTemplate` utilities to the element templates API.

This will make https://github.com/camunda/camunda-bpmn-js/issues/250 easier


nikku:
Let's document them [here](https://github.com/bpmn-io/element-templates/blob/main/docs/API.md) as a follow-up.

src/provider/cloud-element-templates/ElementTemplates.js
```diff
@@ -62,6 +69,55 @@ export default class ElementTemplates extends DefaultElementTemplates {
...
+  }
+
+  /**
+   * Update template to latest version.
+   *
+   * @param {djs.model.Base} element
+   * @param {ElementTemplate} newTemplate
+   * @returns {djs.model.Base} the updated element
+   */
+  updateTemplate(element, newTemplate) {
```
nikku, line 99: The reason we have this (over `apply`) is that we want to tell appart apply and update, right? Fine with me.

nikku: Alternatively we could emit an event in `apply` that allows us to distinguish `apply` and update. Then we have one less API to maintain.

src/provider/cloud-element-templates/ElementTemplatesPropertiesProvider.js
```diff
@@ -44,8 +43,8 @@ export default class ElementTemplatesPropertiesProvider {
         label: 'Template',
         component: createElementTemplatesGroup({
           getTemplateId,
-          unlinkTemplate,
-          updateTemplate
+          unlinkTemplate: this._elementTemplates.elementTemplates,
```
nikku, line 46: Let's make this an arrow function instead? 

```javascript
unlinkTemplate: (...args) => this._elementTemplates.unlinkTemplate(...args)
```

I feel this is easier to understand; also I don't want us to encorage "execute as functions" on instances.

Also, this is likely a typo (cf. `_elementTemplates.elementTemplates`) and not going to work. Please test :slightly_smiling_face:

src/provider/cloud-element-templates/ElementTemplates.js
smbea: So keeping only `applyTemplate` in the API and fire `.update`/`.apply` based on if the templates are the same or not (minus the version)?

src/provider/element-templates/ElementTemplates.js
```diff
@@ -196,6 +203,50 @@ export default class ElementTemplates {
...
+   * Update template to latest version.
+   *
+   * @param {djs.model.Base} element
+   * @param {ElementTemplate} newTemplate
+   * @returns {djs.model.Base} the updated element
+   */
+  updateTemplate(element, newTemplate) {
+    this._eventBus.fire('elementTemplates.update', { element, newTemplate });
+
+    return this.applyTemplate(element, newTemplate);
```
nikku, line 240: Same like above; if it is all just "applying a new template", then we could enhance whatever is emitted in `applyTemplate`, to distinguish apply and update.

src/provider/element-templates/components/ElementTemplatesGroup.js
```diff
@@ -143,24 +136,24 @@ function SectionToggle({ open }) {
...
-    return <UnknownTemplate element={ element } unlinkTemplate={ unlinkTemplate } />;
+    return <UnknownTemplate element={ element } />;
   } else if (templateState.type === 'OUTDATED_TEMPLATE') {
     return (
       <OutdatedTemplate
         element={ element }
         templateState={ templateState }
-        unlinkTemplate={ unlinkTemplate }
-        updateTemplate={ updateTemplate } />
+        unlinkTemplate={ elementTemplates.unlinkTemplate }
```
nikku, line 155: Any reason we don't use the same pattern like with `UnknownTemplate` and `AppliedTemplate` here?

src/provider/cloud-element-templates/ElementTemplatesPropertiesProvider.js
smbea: I know why it didn't. We don't actually need that anymore - will follow up

nikku: Also, as we have the `useService` pattern, we may use it here, too.

src/provider/cloud-element-templates/ElementTemplates.js
nikku: Yea. Or just fire apply (because from the properties panel point of view there is only _apply_, but provide enough context so that integrators can tell apart apply and update.

smbea: I decided to merge the API function but separate the events. We already assume a `update` event with tracking so let's just keep it

nikku: I like that reasoning. :)

src/provider/element-templates/ElementTemplates.js
```diff
@@ -14,14 +14,21 @@ import {
...
 
 /**
  * Registry for element templates.
  */
 export default class ElementTemplates {
-  constructor(commandStack) {
+  constructor(commandStack, replace, selection, eventBus, modeling, injector) {
     this._commandStack = commandStack;
+    this._replace = replace;
+    this._selection = selection;
```
nikku, line 28: We don't seem to use `selection` and `replace`.

src/provider/element-templates/ElementTemplates.js
```diff
@@ -194,8 +206,41 @@ export default class ElementTemplates {
...
     return context.element;
   }
+
+  /**
+   * Remove template from a given element.
+   *
+   * @param {djs.model.Base} element
+   */
+  removeTemplate(element) {
+    removeTemplateHelper(element, this._injector);
```
nikku, line 220: Any reason we don't inline the `removeTemplateHelper` here? Any reason we don't emit `elementTemplate.remove`?

test/spec/provider/element-templates/ElementTemplates.spec.js
```diff
@@ -393,6 +399,149 @@ describe('provider/element-templates - ElementTemplates', function() {
 
   });
 
+
+  describe('removeTemplate', function() {
```
nikku, line 403: :tada: So good!

src/provider/cloud-element-templates/ElementTemplates.js
```diff
@@ -3,17 +3,24 @@ import {
   getTemplateVersion
 } from './Helper';
 
+import {
+  removeTemplate as removeTemplateHelper
```
nikku, line 7: We'd re-use this via re-using the super-class implementation (cf. `DefaultElementTemplate`.

src/provider/cloud-element-templates/ElementTemplates.js
```diff
@@ -50,8 +57,13 @@ export default class ElementTemplates extends DefaultElementTemplates {
    */
   applyTemplate(element, newTemplate) {
 
+    let action = 'apply';
     const oldTemplate = this.get(element);
 
+    if (newTemplate && oldTemplate && (newTemplate.id === oldTemplate.id)) {
+      action = 'update';
```
nikku, line 64: We'd want to fire `remove`, if we un-apply the template (no new template is given)?

src/provider/element-templates/ElementTemplates.js
smbea: The reason is that this is the same logic for platform and cloud. Previously we were using re-exporting this utility in the cloud utilities file. This is so we can keep re-using the same code - easier maintenance.

The event is within the utility too.

src/provider/element-templates/ElementTemplates.js
smbea: This is leftover because the logic that uses this was moved into `removeTemplateHelper`

nikku (changes requested):
Looks great. Found two more things you may want to look into:

* Also insource `templateUtil#removeTemplateHelper`; I do not see a reason we should not.
* Re-use it via super-class.
* Potentially also emit `remove` in applyTemplate

src/provider/element-templates/ElementTemplates.js
nikku: Re-use in class hierarchies is as simple as not implementing the method (in Cloud-element templates).

So I believe we could get rid of the helper entirely, and use inheritance.

nikku: Of course, functional composition works, too.

src/provider/element-templates/ElementTemplates.js
```diff
@@ -194,8 +209,90 @@ export default class ElementTemplates {
...
+
+    const newBusinessObject = createBlankBusinessObject(element, this._injector);
+
+    const newElement = replace.replaceElement(element, {
+      type: type,
+      businessObject: newBusinessObject,
+      eventDefinitionType: eventDefinitionType
+    });
+
+    selection.select(newElement);
```
nikku, line 242: Could you verify if this is necessary? We handle selection after replace in diagram-js now, cf. https://github.com/bpmn-io/diagram-js/blob/develop/lib/features/replace/ReplaceSelectionBehavior.js#L3.

smbea: Yup, we don't need this. I removed the selection here

src/provider/element-templates/ElementTemplates.js
```diff
@@ -194,8 +209,87 @@ export default class ElementTemplates {
...
+    eventBus.fire('elementTemplates.remove', { element });
+
+    const businessObject = getBusinessObject(element);
+
+    const type = businessObject.$type,
+          eventDefinitionType = getEventDefinitionType(businessObject);
+
+    const newBusinessObject = createBlankBusinessObject(element, this._injector);
+
+    replace.replaceElement(element, {
```
nikku, line 235: Let's return the new element. Allows us to write actual tests.

src/provider/element-templates/ElementTemplates.js
```diff
@@ -194,8 +209,87 @@ export default class ElementTemplates {
...
+    });
+  }
+
+  /**
+   * Unlink template from a given element.
+   *
+   * @param {djs.model.Base} element
+   */
+  unlinkTemplate(element) {
+    this.applyTemplate(element, null);
```
nikku, line 248: Cf. https://github.com/bpmn-io/bpmn-js-properties-panel/pull/935/files#r1224061460. Let's always return the new or changed element.

src/provider/element-templates/ElementTemplates.js
```diff
@@ -194,8 +209,87 @@ export default class ElementTemplates {
 
     this._commandStack.execute('propertiesPanel.camunda.changeTemplate', context);
 
+    this._eventBus.fire(`elementTemplates.${action}`, payload);
+
     return context.element;
```
nikku, line 214: This is where we already return the "changed element" :)

Cf. https://github.com/bpmn-io/bpmn-js-properties-panel/pull/935/files#r1224061460.

nikku (changes requested):
Looks great. As a final clean-up, let's return the changed element from the API (+ update the tests).

Great cleanup + improvement all together :1st_place_medal:

barmac:
Happy to see this merged 🎉

PR #937, opened 2023-06-13
Remove element templates

marstamm:
They now live in https://github.com/bpmn-io/bpmn-js-element-templates

related to https://github.com/camunda/camunda-modeler/issues/3357


smbea:
So we still consider element templates for rendering purposes (such as https://github.com/bpmn-io/bpmn-js-properties-panel/blob/remove-element-templates/src/render/PanelHeaderProvider.js#L71) but logic is completely removed, right?

marstamm:
Yes, some aspects of element templates can not be easily isolated from the module. Goal for this PR is to remove the main element-template logic from the properties-panel. Individual Providers might still react and use the element template service when it is present

smbea:
Makes sense, just wanted to clarify!

smbea (approved):
🚀

PR #947, opened 2023-07-27
Update properties panel peer dependency

smbea:

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "zeebe-bpmn-moddle": "^0.19.0"
   },
   "peerDependencies": {
-    "@bpmn-io/properties-panel": ">= 2.2",
+    "@bpmn-io/properties-panel": ">= 3.2",
     "bpmn-js": ">= 11.5",
     "camunda-bpmn-js-behaviors": ">= 0.4",
     "diagram-js": ">= 11.9"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -114,7 +114,7 @@
     "zeebe-bpmn-moddle": "^0.19.0"
   },
   "peerDependencies": {
-    "@bpmn-io/properties-panel": ">= 2.2",
```
barmac, line 117: Does it not work with 2.x anymore? If so, that should be marked in breaking changes...

smbea: Nothing breaks. But the new tooltips feature doesn't work, and since some descriptions were migrated to tooltips, then "nothing" appears. I already released these as a breaking change in version v4

barmac: OK so we basically fix the peer dependency change which was already released as a breaking change with v4. Potentially we could also include that in the breaking changes section in CHANGELOG. Still, that shouldn't be a problem for users, especially as the release is so recent.

barmac (approved):

package.json
smbea: Mention that it needs `@bpmn-io/properties-panel": ">= 3.2` now? I noticed we've never mentioned this in "breaking changes"

PR #958, opened 2023-08-24
Remove properties panel css

marstamm:
This helps us streamline our release process and prevents css mismatches in downstream libraries.

Related Slack discussion: https://camunda.slack.com/archives/GP70M0J6M/p1692870672157409


nikku (approved):
Looks good. I think this more clearly separates things; and if you want a "bundle", then camunda-bpmn-js or other projects are your go-to place.

PR #967, opened 2023-09-12
chore(CHANGELOG): add FEEL popup menu entry

nikku:
This adds a previously "under the radar" major addition.


philippfromme (approved):
There it is! 🏁

Issue #981, opened 2023-09-27
Redo does not work in FEEL input when undone till empty input


barmac:
### Describe the Bug

Given I type in the FEEL entry, when I press cmd+Z and cmd+Y, in FEEL input, redo stops to work as soon as I reach empty input with undo.

### Steps to Reproduce

1. Type in FEEL input
2. Undo until input is empty
3. Try to redo
4. Nothing happens

### Expected Behavior

Redo should work.

### Environment

- Host (Browser/Node version), if applicable: [e.g. MS Edge 18, Chrome 69, Node 10 LTS]
- OS: [e.g. Windows 7]
- Library version: Confirmed in v5.3.0-v5.5.0

marstamm:
I can only reproduce this issue in Input/Output fields. Other FEEL enabled fields, such as Task type or Script Tasks FEEL expressions manage the undo stack correctly.

PR #983, opened 2023-10-10
fix(Input/Output): keep undo/redo stack

marstamm:
This PR removes the empty `source` value for newly created Input/Output parameters. I decided to go with the convention "no value => remove property" that we use in other places, such as Task definition type. I verified that the deployment error in zeebe is the same for empty value and missing property.

I added integration tests to verify the scenario described in #981.

Recording 2023-10-10 at 10 59 16


nikku (commented):
Great stuff! :cake:

nikku (approved):

PR #984, opened 2023-10-11
feat(FEEL): use moddle element to filter for variables 

marstamm:
We added this functionality to the variable resolver a while back but never properly integrated it. Ooops. Better late than never I guess. 

image

related to https://github.com/camunda/camunda-modeler/issues/3510

### DIFF
--- a/src/provider/HOCs/withVariableContext.js
+++ b/src/provider/HOCs/withVariableContext.js
@@ -2,15 +2,9 @@ import { getVariablesForElement } from '@bpmn-io/extract-process-variables/zeebe
 import { useEffect, useState } from '@bpmn-io/properties-panel/preact/hooks';
 import { useService } from '../../hooks';
 
-function useServiceIfAvailable(service, fallback) {
-  const resolved = useService(service, false);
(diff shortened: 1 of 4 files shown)


src/provider/HOCs/withVariableContext.js
```diff
@@ -53,4 +43,20 @@ export function withVariableContext(Component) {
...
+  if (!resolved) {
+    return fallback;
+  }
+
+  return resolved;
+}
+
+const fallbackResolver = {
+  getVariablesForElement: bo => getVariablesForElement(bo)
+};
```
barmac, line 62: This made me wonder if it's not a bug. `const` and `let` variables are not hoisted, so theoretically if one calls `withVariableContext` immediately after import, the variable may throw error of access before initialisation. I guess it's not a problem since the CI is passing, but I'd rather expect such defaults to be declared on top of the file.

barmac: I will do it.

barmac (approved):

PR #985, opened 2023-10-13
House keeping

nikku:
Usual dependency updates, and dropping unused + deprecated `@testinglibrary/preact-hooks` dependency.


barmac:
This fails badly in the CI. Looks like `preact` is missing from the legacy `@bpmn-io/properties-panel`.

barmac:
It is installed though: https://unpkg.com/browse/@bpmn-io/properties-panel@3.10.0/preact/

nikku:
Reverted the `mjs` rename that caused this - https://github.com/bpmn-io/properties-panel/pull/305.

barmac:
This will make sure that we know immediately if the problem comes from integration deps or not: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/987

nikku:
Should be fixed via upstream `v3.10.1` release.

barmac (approved):

PR #986, opened 2023-10-13
chore: export module using `mjs` extension

nikku:
Following up on https://github.com/bpmn-io/properties-panel/pull/303.

This allows module bundlers to properly interpret the file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "node": "*"
   },
   "main": "dist/index.js",
-  "module": "dist/index.esm.js",
+  "module": "dist/index.mjs",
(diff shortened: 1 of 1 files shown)


barmac (approved):
I think this still makes sense because `@bpmn-io/properties-panel` failure was due to unique structure of the library, which doesn't occur here.

nikku:
Let's park this for now regardless.

nikku:
We cannot do this or we'll break downstream integrations, cf.:

image

If we ever exposed an `.mjs` file, then we'd need to ensure that dependencies are also modules, and/or expose the module exports we search for :see_no_evil:.

PR #987, opened 2023-10-13
ci: continue if peer deps job fails

barmac:
This aims to allow us to discover quickly whether a CI failure is caused by incompatibility or another problem.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -2,17 +2,17 @@ name: CI
 on: [ push, pull_request ]
 jobs:
   Build:
-
+    runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.integration }}
     strategy:
       matrix:
-        os: [ ubuntu-latest ]
-        node-version: [ 20 ]
+        integration: [ true ]
         integration-deps:
         - diagram-js@11.9 bpmn-js@11.5
         - "@bpmn-io/properties-panel@3"
-        - "" # as defined in package.json
-
-    runs-on: ${{ matrix.os }}
+        include:
+          - integration-deps: "" # as defined in package.json
+            integration: false
 
     steps:
     - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -2,25 +2,25 @@ name: CI
...
+        integration: [ true ]
         integration-deps:
         - diagram-js@11.9 bpmn-js@11.5
         - "@bpmn-io/properties-panel@3"
-        - "" # as defined in package.json
-
-    runs-on: ${{ matrix.os }}
+        include:
+          - integration-deps: "" # as defined in package.json
+            integration: false
```
nikku, line 15: What does this do? I don't see `integration` being used.

barmac: `integration` is a notion that a job is part of the `integration-deps` suite. This is used to determine whether other jobs should continue if the job fails via `continue-on-error`: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/987/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R6

barmac: Alternatively, we could call this property `continue-on-error` but I didn't want to mix up GHA keywords and our custom properties. If it makes more sense, I will rename this.

barmac:
BTW with this settings we can see clearly that the latest release of `@bpmn-io/properties-panel` causes the failure.

.github/workflows/CI.yml
```diff
@@ -2,25 +2,25 @@ name: CI
 on: [ push, pull_request ]
 jobs:
   Build:
-
+    runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.integration }}
```
nikku, line 6: This is the trick, thanks!

nikku (approved):
Makes sense. Cool stuff.

Issue #1011, opened 2024-01-04
Properties Panel Header Displays Element Template Icon Found in Template Instead of XML


zxuanhong:
### Is your feature request related to a problem? Please describe.

1. The properties at the top of the icon with ElementTemplateIconRenderer get icon logically consistent. Get it directly from the element or read it from the built-in icon.
2. If the logic is consistent, you can customize the icon of an element
3. Key code ElementTemplateIconRenderer access icon
<img width="843" alt="image" src="https://github.com/bpmn-io/bpmn-js-properties-panel/assets/62740686/78ba2a38-44cf-48ab-894a-65baafaffce4">
4. And now once the template is lost, it will also cause the element icon to render properly, but the properties header icon will not render properly.For example
<img width="1135" alt="image" src="https://github.com/bpmn-io/bpmn-js-properties-panel/assets/62740686/8add2097-687d-4448-984d-7a762454e55c">

nikku:
@zxuanhong We read it from the rendered XML because we want the icon to be present also in the case of missing element template.

Do I understand you correctly that your feature request is to use the same mechanism to read the icon in the properties panel? There we currently read the icon from the connected element template.

zxuanhong:
@nikku yes. this pr will solve this problem #1012

PR #1012, opened 2024-01-04
Properties Panel Header Displays Element Template Icon Found in XML

zxuanhong:
1. modify the rendering panel header icon to match the element template icon renderer. Even if the template is not loaded, it will not affect icon rendering.for example(not loaded templates)：
iShot_2024-01-04_19 07 11
iShot_2024-01-04_19 07 25

2. can indirectly implement custom icons for element

----

Closes https://github.com/bpmn-io/bpmn-js-properties-panel/issues/1011


src/render/PanelHeaderProvider.js
```diff
@@ -191,4 +188,10 @@ function getTemplateDocumentation(element, elementTemplates) {
   const template = getTemplate(element, elementTemplates);
 
   return template && template.documentationRef;
+}
+
+function getModelerTemplateIcon(element, iconProperty) {
+  iconProperty = iconProperty || 'zeebe:modelerTemplateIcon';
```
philippfromme, line 194: Where does `iconProperty` come from? Since `PanelHeaderProvider` is generic it doesn't make sense to assume `zeebe:modelerTemplateIcon` as the default.

philippfromme (changes requested):
Thank you for your contribution. I've left a remark.

src/render/PanelHeaderProvider.js
zxuanhong: @philippfromme iconProperty come from this
image

zxuanhong: @philippfromme @nikku Have been corrected
iShot_2024-01-04_21 23 54
iShot_2024-01-04_21 22 41

src/render/PanelHeaderProvider.js
```diff
@@ -66,58 +66,55 @@ export function getConcreteType(element) {
...
+    },
 
-    if (elementTemplates) {
-      const template = getTemplate(element, elementTemplates);
+    getElementIcon: (element) => {
+      const concreteType = getConcreteType(element);
 
-      if (template && template.icon) {
-        return () => <img class="bio-properties-panel-header-template-icon" width="32" height="32" src={ template.icon.contents } />;
+      const templateIcon = getModelerTemplateIcon(element, iconProperty);
```
philippfromme, line 90: Instead of passing `iconProperty` all the way down you can use the `useService` (which was also used previously) hook to get hold of it:

```
const config = useService('config.elementTemplateIconRenderer', false);

if (config) {
  ...
}
```

zxuanhong: @philippfromme Thank you very much for the reminder.It looks perfect now😄

src/render/PanelHeaderProvider.js
```diff
@@ -191,4 +187,12 @@ function getTemplateDocumentation(element, elementTemplates) {
   const template = getTemplate(element, elementTemplates);
 
   return template && template.documentationRef;
+}
+
+function getModelerTemplateIcon(element) {
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const config = useService('config.elementTemplateIconRenderer', false);
+  const { iconProperty = 'zeebe:modelerTemplateIcon' } = config || {};
```
philippfromme, line 195: As this component is generic, we don't want to assume `zeebe:modelerTemplateIcon` as a default here. If there is no configured `iconProperty`, we simply want to fall back to the original icon.

philippfromme (changes requested):

src/render/PanelHeaderProvider.js
philippfromme: Much better. 😉

src/render/PanelHeaderProvider.js
zxuanhong: I made the judgment here. Because the default value on the icon template is 'zeebe:modelerTemplateIcon'.And I won't set the icon template config for normal use.
image

zxuanhong: Are you sure you want to remove the default values???

nikku: @philippfromme On the other hand it wants to play nicely with our other utilities (i.e. the icons renderer). There we assume a default.

Hence I think it is safe to keep it hard coded per default _or_ get the icon from the icon renderer (if it exists).

philippfromme: Okay, makes sense. We can go ahead with it then. 👍🏻

philippfromme:
Merged through https://github.com/bpmn-io/bpmn-js-properties-panel/commit/c81dc4fd4e4be63bf189e471d30da1cdf9f442a7. I  squashed the commits and adjusted the tests. Thank you for your contribution. 🎖️

PR #1013, opened 2024-01-10
test: verify properties panel works in read-only mode

nikku:
_Context: https://github.com/camunda/web-modeler/issues/7623._

What is left is three issues:

* Custom styling (marking fields as disabled)
* Disabling form element focus (may need to be solved in properties panel)
* Hiding (+) buttons

---

For the readonly properties panel case I _could_ imagine to build dedicated support into the core:

```javascript
const modelerWithReadonlyPropertiesPanel = new BpmnModeler({
  propertiesPanel: {
    readonly: true
  }
});
```

This would do the following:

* mark inputs and checkboxes as _readonly_
* do not show add markers for sections

...and would in the process solve the issue of editing and focus on the HTML level.


nikku:
@nithinssabu the results of our exploration.

nithinssabu:
@nikku If I try to dynamically attach the properties panel like this,
```js
const propertiesPanel = modeler.get('propertiesPanel');
propertiesPanel.attachTo(propertiesContainer);
```

getting the following error,
image

Can you please have a look? I spent quite some time on it.

nikku:
@nithinssabu For context, please share the web modeler issue you're currently working on.

philippfromme:
We already have [styles for `disabled="true"`](https://github.com/bpmn-io/properties-panel/blob/441d2e4c3e6b401e91095cbce5832137f843b037/src/assets/properties-panel.css#L520C29-L520C38) inputs, we're just not setting that attribute which we could based on the configuration.

nithinssabu:
> For context, please share the web modeler issue you're currently working on.

https://github.com/camunda/web-modeler/issues/7623

nithinssabu:
@nikku Can I get permission to push changes to this branch?

nithinssabu:
For my use case, I also need to get linting to work in the read-only mode. Currently getting the following error:

```js
Error: No provider for "elementTemplates"! (Resolving: elementTemplates -> elementTemplates -> elementTemplates)
    at error (index.esm.js:136:1)
    at Object.get (index.esm.js:122:1)
    at NavigatedViewer.get (index.esm.js:176:1)
    at createModeler (BpmnPropertiesPanelRenderer.readonly.spec.js:142:35)
    at async Context.eval (BpmnPropertiesPanelRenderer.readonly.spec.js:177:20)
```

I pushed a commit reproducing this in the current test case. Can you please help me workaround this problem?

nithinssabu:
> What is left is three issues:
> 
> * Custom styling (marking fields as disabled)
> * Disabling form element focus (may need to be solved in properties panel)
> * Hiding (+) buttons

I was able to solve these using CSS and for the keyboard focus, using Javascript to set the `readonly` attribute on the input elements. Also, used `MutationObserver` to set this when the panel is re-drawn. https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1013/commits/d8563b99ddb6ca0fc97880b2a8683b1336525815

https://github.com/bpmn-io/bpmn-js-properties-panel/assets/5987816/bbbf57f8-725f-4cfe-9d52-72e8a76e8a6b

@philippfromme I used `readonly` attribute instead of `disabled` because with `disabled`, it was not possible to tab through the elements. The user may want to focus the inputs and may be copy the contents.

nikku:
Web modeler shipped without core support. Moving to `backlog` to track future improvement options.

nikku:
This has no urgency right now, as web modeler was able to work around this issue. Could still be an enabler for the purpose of "understanding" diagrams, in read only contexts.

PR #1016, opened 2024-01-28
feat(conditional-variable-events): present event options in tooltip

timstawowski:
Show all possible variable event values to avoid searching for them.

### DIFF
--- a/src/provider/camunda-platform/properties/ConditionProps.js
+++ b/src/provider/camunda-platform/properties/ConditionProps.js
@@ -421,14 +421,26 @@ function VariableEvents(props) {
     });
   };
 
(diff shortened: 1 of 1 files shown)


barmac:
Hi,

thanks for your contribution! The initial solution didn't work, but I managed to fix that.

image

I merged your contribution via https://github.com/bpmn-io/bpmn-js-properties-panel/commit/4e3684bb505c605834d19abbcbbc8af31ff5e5d5. It occurred that the `tooltip` prop is expecting a preact element and not a component. I will create an issue for this.

timstawowski:
Hi @barmac,

thanks for the fix and sorry for presenting without local testing.

PR #1022, opened 2024-02-01
fix(Id): create validation function as callback

marstamm:
Root cause: not properly using Callback hooks, changing the Validation function every render. This caused a recheck with the correct value, since the incorrect one is not applied to the XML.

### DIFF
--- a/src/provider/bpmn/properties/IdProps.js
+++ b/src/provider/bpmn/properties/IdProps.js
@@ -5,6 +5,8 @@ import {
 
 import { TextFieldEntry, isTextFieldEntryEdited } from '@bpmn-io/properties-panel';
 
+import { useCallback } from '@bpmn-io/properties-panel/preact/hooks';
+
 import {
   useService
 } from '../../../hooks';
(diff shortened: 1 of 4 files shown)


src/provider/bpmn/properties/IdProps.js
```diff
@@ -50,15 +52,15 @@ function Id(props) {
     });
   };
 
-  const getValue = (element) => {
+  const getValue = useCallback((element) => {
     return element.businessObject.id;
-  };
+  }, []);
```
nikku, line 57:
```suggestion
  }, [ element ]);
```

src/provider/bpmn/properties/IdProps.js
```diff
@@ -50,15 +52,15 @@ function Id(props) {
     });
   };
 
-  const getValue = (element) => {
+  const getValue = useCallback((element) => {
     return element.businessObject.id;
```
nikku, line 56:
```suggestion
    return getBusinessObject(element).id;
```

nikku (approved):
Great stuff. Thanks for the swift follow-up!

PR #1026, opened 2024-02-20
Support `zeebe:userTask`

barmac:
Requires https://github.com/camunda/camunda-bpmn-js-behaviors/pull/67 to be integrated.

Related to https://github.com/camunda/camunda-modeler/issues/4087


barmac:
The external reference field should be FEEL-optional

barmac:
In https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1026/commits/0540c55c4a9a40d839cf406536e0fd99591c46ac, I am removing the tests which caused CI to fail. We don't want to remove output parameters via "propagateChildVariables" switch anymore.

nikku:
To consider: Change `implementation` which is duplicated consistently to `Type`:

image

:arrow_down: 

Type for these kinds of choices is already a mentioned term.

image

nikku (approved):
Works like a charm. I like how we retain the existing context where appropriate, too.

Consider one feedback (remove double implement) as a potential follow-up.

barmac:
I updated the label and will be merging this shortly.

PR #1046, opened 2024-04-23
Add called decision guidance

### LINKED ISSUE #1039: Link to the documentation in Called Decision group
### What should we do?

Add tooltip with a link to https://docs.camunda.io/docs/components/modeler/bpmn/business-rule-tasks/#defining-a-called-decision in the called decision.

### Why should we do it?

Some users get confused about this section, cf. https://forum.camunda.io/t/search-resource-in-practice-for-beginner/51423/8?u=crobbins215

nikku:
screenshot 0PuZAM

screenshot ktF6yJ

Closes #1039

### DIFF
--- a/src/contextProvider/zeebe/TooltipProvider.js
+++ b/src/contextProvider/zeebe/TooltipProvider.js
@@ -42,7 +42,7 @@ const TooltipProvider = {
 
     return (
       <div>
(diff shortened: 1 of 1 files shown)


barmac:
Great addition!

PR #1047, opened 2024-04-26
Remove unsupported flag "shouldSort"

AlexanderSkrock:
This pull request aims to fix [#311 in properties-panel](https://github.com/bpmn-io/properties-panel/issues/311) together [#353 in properties-panel](https://github.com/bpmn-io/properties-panel/pull/353). Currently, it is being developed.

- [x] Remove "shouldSort" property where ever the flag was set to `false` explicitly
- [x] Implement sorting where necessary
       => Potentially every component which did not explicitly set `shouldSort` to `false`, but we should use  this opportunity to evaluate where it actually should be sorted.
- [x] Write a note in CHANGELOG.md

Regarding the `shouldSort` flag, I saw the following throughout this project.

**Camunda 7 - Camunda Platform**
Usages without the `shouldSort` flags, thus using the previous default of sorting:
- FieldInjectionGroup
- InMappingGroup
- OutMappingGroup
- ExecutionListenerGroup
- TaskListenerGroup
- ConnectorInputGroup
- ConnectorOutputGroup
- ExtensionPropertiesGroup

Usages with the `shouldSort` flag set to `false`, thus already disabling sorting:
- ErrorsGroup
- FormDataGroup
- InputGroup
- OutputGroup
- ProcessVariablesGroup
- ExtensionPropertiesGroup

**Camunda 8 - Zeebee**
Usages with the `shouldSort` flag set to `false`, thus already disabling sorting:
- ExtensionPropertiesGroup
- HeaderGroup
- InputGroup
- OutputGroup

----

Related to https://github.com/bpmn-io/properties-panel/issues/311

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to [bpmn-js-properties-panel](https://github.com/bpmn-io/bpm
 
 ___Note:__ Yet to be released changes appear here._
 
+* `FEAT`: drop alphabetic sorting of list entries  ([#1047](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1047))
+
+
 ## 5.16.0
 
 * `FEAT`: add hint for the process ID field in the Camunda 7 ([#1038](https://github.com/bpmn-io/bpmn-js-properties-panel/issues/1038))
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "@babel/plugin-transform-react-jsx": "^7.23.4",
         "@bpmn-io/element-template-chooser": "^1.0.0",
         "@bpmn-io/element-templates-icons-renderer": "^0.3.0",
-        "@bpmn-io/properties-panel": "^3.18.2",
+        "@bpmn-io/properties-panel": "^3.19.0",
         "@bpmn-io/variable-resolver": "^1.2.2",
         "@rollup/plugin-alias": "^5.1.0",
         "@rollup/plugin-babel": "^6.0.4",
@@ -622,9 +622,9 @@
       }
     },
     "node_modules/@bpmn-io/properties-panel": {
-      "version": "3.18.2",
-      "resolved": "https://registry.npmjs.org/@bpmn-io/properties-panel/-/properties-panel-3.18.2.tgz",
-      "integrity": "sha512-IQ6NUZ4McSmr6KLyptnhnKxBind5Oz+FSZ5u8MJX/s/10RRj+RIVYCBS2UnfCKHZCE9YMWTdCHdA7XQ4lIjuzw==",
+      "version": "3.19.0",
+      "resolved": "https://registry.npmjs.org/@bpmn-io/properties-panel/-/properties-panel-3.19.0.tgz",
+      "integrity": "sha512-ZVr8LC8ZiMYRkh9Vf4EnuQ7hJIzjBLpLUoex+q5/tGj1I2B/iFAgDUaqnD/yOyBzkRDi1K9IedsUA01XaqG13w==",
       "dev": true,
       "dependencies": {
         "@bpmn-io/feel-editor": "^1.3.0",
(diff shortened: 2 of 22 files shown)


AlexanderSkrock:
Hi @nikku ,
it would be great if you could have a look as well regarding the question which components should keep alphabetical sorting. Thank you in advance.

nikku:
@AlexanderSkrock I can confirm that non of the fields you mentioned should be sorting.

We can argue about "pure visual only", such as C7 process variables; but then again we can sort on the business logic level.

AlexanderSkrock:
Seems like my changes from `properties-panel` were not used. I did one more little fix there for the change dot appearing too late, but I can not confirm the remaining failures locally.

nikku (dismissed):
Auto-focus seems to need adjustment in different places (I sketched the change on one example on top of this PR); `List#compareFn` could be dropped (I pushed that change on top of this PR).

Cf. https://github.com/bpmn-io/properties-panel/pull/353#issuecomment-2096326398 for details.

AlexanderSkrock:
Thank you for having a look! I tried to follow up on your changes regarding auto-focus, but for the `ListProps` it seems non-trivial as there is no `data-entry-id` yet, I could hook into.

I wanted to add `autoFocusEntry: `[data-entry-id="${id}-listItem-${items.length - 1}"] input` because that is how its done for all other components, but then I'd need to set the `data-entry-id` on a very high level, e. g.  [here[(https://github.com/AlexanderSkrock/bpmn-js-properties-panel/blob/2a42d182b007ab9e2aafe5d2fa55a83b657ff57d/src/provider/camunda-platform/properties/ListProps.js#L119).

Otherwise if a `data-entry-id` with a `value`-suffix would be okay such as `"${id}-listItem-${items.length - 1}-value"`, then we could incorporate this in the `SimpleEntry` component in `properties-panel` core project, e. g. [here](https://github.com/AlexanderSkrock/properties-panel/blob/6788c4cee6fd5fc7cc7f3c73f87f7085638f0900/src/components/entries/Simple.js#L53). Btw, I wondered why we add a prefix to the id there, which we don't for all other entries.

In addition to both of these approaches with the current `SimpleEntry` component, I could imagine to simply switch to `TextField` for entries of `ListProps`, which should work out of the box in combination with `autoFocusEntry: `[data-entry-id="${id}-listItem-${items.length - 1}"] input`.

Would be great, if you could have a look here again, @nikku , as I am not sure which is the best approach. While the last one mentioned seems best to me at first glance, it might have non-obvious changes I am not able to understand in depth.

nikku:
@AlexanderSkrock I'll only be able to follow-up on your comments next week.

nikku:
@AlexanderSkrock I was able to review your changes, great work! The only thing missing is the auto focus which is now broken in some cases. 

Was not fully able to grasp some of your thoughts in https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1047#issuecomment-2097993831. I'd propose to make auto-focus work in the most simple way :upside_down_face:. This implies:

* Use the `data-entry-id` hack for nested lists that return a specific component
* Use simple entry addressing `id + '-sub-entry-key'` for flat lists (https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1047/commits/db441b88aba48cb7ec1d54a38345feb77d5ca620, https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1047/commits/dc09c9e97a6abedf1b65c5b0e9a9cd2d80d82a9e)

Of course we could refactor further; but let's take things one step at a time.

nikku:
To be clear: I regard `data-entry-**` auto-focussing as a hack, in a perfect world we'd refactor the list entry to allow passing a dedicated sub-entry property (just a suffix), or provide other resolutions to the common "simply define what to auto-focus" problem:

* We could define an implicit auto-focus, without any explicit configuration - entries can be tagged, and upon creation / "focus" the first entry that is tagged as "could have auto-focus" receives it.
* We could rework the simple and list entries to be able to pass focus in another way.

But then again, looking at the small amount of changes needed here I'm fine to accept the contribution as is.

nikku:
I'll look into rebasing this. Thanks for your work on this issue :clap:

nikku:
Updated to build on top of `@bpmn-io/properties-panel@3.19.0`.

nikku:
I can reproduce the test single failure locally and am investigating.

nikku:
Addressed everything mentioned in https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1047#issuecomment-2124706979. Waiting for CI to pass.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to [bpmn-js-properties-panel](https://github.com/bpmn-io/bpm
 
 ___Note:__ Yet to be released changes appear here._
 
+* `FIX`: removed sorting on list groups as it may affect semantics ([#1047](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1047))
```
nikku, line 9:
```suggestion
* `FEAT`: drop alphabetic sorting of list entries  ([#1047](https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1047))

```

nikku (approved):

nikku:
Thanks for your contribution @AlexanderSkrock. We really apprechiate it :medal_sports: :clap:

PR #1055, opened 2024-05-28
fix: display process name only once for C7

barmac:
Related to https://github.com/camunda/camunda-modeler/issues/4317


nikku:
@barmac Let's rebase this on `main`. `v5.17.0` of the properties panel is out.

nikku:
Released as https://github.com/bpmn-io/bpmn-js-properties-panel/releases/tag/v5.17.1

PR #1072, opened 2024-08-14
feat: support editing `zeebe:priorityDefinition`

### LINKED ISSUE #1069: Support editing priority in Zeebe
### Is your feature request related to a problem? Please describe.

As part of feature parity, we are adding a way to add a simple priority on user tasks to sort what's important from what isn't. This new priority extension element will be available on user tasks, and so we need to also provide the UI for it. 

Sample XML:

```xml
<userTask id="" name="">
  <bpmn:extensionElements>
    <zeebe:priorityDefinition priority="15" />
  </bpmn:extensionElements>
</userTask> 
```

### Describe the solution you'd like

Field that allows setting a value from 0-100 or a feel expression as per the product hub epic.
Add some validation to guide the user to enter the type.

* [x] New priority field is added in Assignment
* [x] Priority is an integer per default
* [x] Priority can be optionally a FEEL expression
* [x] A documentation tooltip with a link to docs is added

### Describe alternatives you've considered

n/a

### Additional context

Product hub epic: https://github.com/camunda/product-hub/issues/217
Root: https://github.com/camunda/camunda/issues/19833
Moddle issue: https://github.com/camunda/zeebe-bpmn-moddle/issues/62

Related to https://github.com/camunda/camunda-modeler/issues/4450

Skaiir:
### Proposed Changes

Adds an entry in the properties panel to edit the `zeebe:priorityDefinition::priority` of a user task. Should have a tooltip with some info and a (currently empty) link to docs. 

Closes #1069

image

`npx @bpmn-io/sr bpmn-io/bpmn-js#main -l bpmn-io/bpmn-js-properties-panel#1069-support-priority-editing`

### DIFF
--- a/src/provider/zeebe/properties/PriorityDefinitionProps.js
+++ b/src/provider/zeebe/properties/PriorityDefinitionProps.js
@@ -1,0 +1,155 @@
+import {
+  getBusinessObject,
+  is
+} from 'bpmn-js/lib/util/ModelUtil';
+
+import { isFeelEntryEdited } from '@bpmn-io/properties-panel';
+
+import {
+  getExtensionElementsList
+} from '../../../utils/ExtensionElementsUtil';
+
+import {
+  createElement
+} from '../../../utils/ElementUtil';
+
+import {
+  useService
+} from '../../../hooks';
+
+import { FeelEntryWithVariableContext } from '../../../entries/FeelEntryWithContext';
+
+
+export function PriorityDefinitionProps(props) {
+  const {
+    element
+  } = props;
+
+  if (!is(element, 'bpmn:UserTask')) {
+    return [];
+  }
+
+  return [
+    {
+      id: 'priorityDefinitionPriority',
+      component: Priority,
+      isEdited: isFeelEntryEdited
+    }
+  ];
+}
+
+function Priority(props) {
+  const {
+    element
+  } = props;
+
+  const commandStack = useService('commandStack');
+  const bpmnFactory = useService('bpmnFactory');
+  const translate = useService('translate');
+  const debounce = useService('debounceInput');
+
+  const getValue = () => {
+    return (getPriorityDefinition(element) || {}).priority;
+  };
+
+  const setValue = (value) => {
+    const commands = [];
+
+    const businessObject = getBusinessObject(element);
+
+    let extensionElements = businessObject.get('extensionElements');
+
+    // (1) ensure extension elements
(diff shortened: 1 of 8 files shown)


philippfromme:
Thanks for the contribution @Skaiir! Will have a look! 🫶

src/provider/zeebe/properties/PriorityDefinitionProps.js
```diff
@@ -0,0 +1,162 @@
...
+          }
+        }
+      });
+    }
+
+    // (4) commit all updates
+    commandStack.execute('properties-panel.multi-command-executor', commands);
+  };
+
+  const tooltip = <div>
```
philippfromme, line 137: We keep the tooltips in the respective [provider](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/main/src/contextProvider/zeebe/TooltipProvider.js). Also, the link doesn't point anywhere at the moment. I guess that's because the documentation doesn't exist yet?

philippfromme (changes requested):

src/provider/zeebe/properties/PriorityDefinitionProps.js
Skaiir: Hey there, yeah it's because it doesn't exist yet. Raised a follow-up here: https://github.com/bpmn-io/bpmn-js-properties-panel/issues/1073.

Skaiir: On a related topic, there will come a time where this method of translating won't be enough. Not being able to move divs around relative to each-other prevents things like links or formatting inside of sentences.

I think something similar to the [Trans component](https://react.i18next.com/latest/trans-component) of react-i18next might make sense. Just floating the idea around.

Skaiir:
Made the adjustments, removed the link entirely for now, follow up here: https://github.com/bpmn-io/bpmn-js-properties-panel/issues/1073.

src/provider/zeebe/properties/PriorityDefinitionProps.js
nikku: > I think something similar to the [Trans component](https://react.i18next.com/latest/trans-component) of react-i18next might make sense. Just floating the idea around.

Let's keep things simple and not over-engineer. Quoting from the site you mentioned:

image

philippfromme (approved):
Great contribution! 🚀

PR #1097, opened 2024-12-11
chore: update terminology to "camunda user task"

Skaiir:
Related to https://github.com/camunda/camunda-modeler/issues/4749


barmac:
We need to adjust the tooltips yet:

image

I will do that.

PR #1109, opened 2025-02-19
fix: subscription coorelation key tooltip added

abdul99ahad:
Related to camunda-modeler: [#4833](https://github.com/camunda/camunda-modeler/issues/4833)

### Proposed Changes

 

Added tooltip for Subscription correlation key

Screenshot 2025-02-19 at 16 54 43

### Checklist

### DIFF
--- a/src/contextProvider/zeebe/TooltipProvider.js
+++ b/src/contextProvider/zeebe/TooltipProvider.js
@@ -361,6 +361,19 @@ const TooltipProvider = {
       </div>
     );
   },
+  'messageSubscriptionCorrelationKey': (element) => {
+    const translate = useService('translate');
+
+    return (
+      <div>
+        <p>
+          { translate('Identifier computed from the process context that is used to correlate an incoming message (e.g. ')}<code>= orderId</code>{translate('). ')}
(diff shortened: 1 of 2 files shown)


nikku:
Not sure if the message brings the point across, what about the following instead:

> Key computed from the process context that is used to correlate an incoming message, e.g. `= orderId`.

abdul99ahad:
> Key computed from the process context that is used to correlate an incoming message

Yeah, that sounds about right. If we want to add a little more detail (maybe redundant) we can do this
> Key computed from the process context is used to correlate an incoming message ensuring that a message is associated with the correct process instance.

not sure about this though

nikku:
I'll keep it short and simple, and go from there.

barmac:
As short as I can get it:
> Expression result is used to correlate a message with a process instance, e.g. `=orderId`.

nikku:
> Expression result is used to correlate a message with a process instance, e.g. =orderId.

My critique is that the result is not used, but instead the result creates an `identifier` that is used to correlate incoming messages. So most appropriate would be:

> Identifier computed from the process context that is used to correlate an incoming message, e.g. `= orderId`.

barmac:
Sounds good to me!

abdul99ahad:
Tooltip content changed. 
Screenshot 2025-02-20 at 12 43 17

barmac (approved):

nikku (approved):

nikku:
Good stuff :clap:

PR #1111, opened 2025-02-21
Fix broken tests

barmac:
### Proposed Changes

This fixes some tests which used to throw uncaught promise exceptions.

I propose also to stop testing for the edited dot existence. These assertions are super flaky and slow (`wait 50 ms` included), and a rewrite to `waitFor` did not resolve flakiness on my machine. I don't think this is a crucial feature, and we tested for it only in some places in the code base.

### Checklist

### DIFF
--- a/test/spec/provider/bpmn/EscalationProps.spec.js
+++ b/test/spec/provider/bpmn/EscalationProps.spec.js
@@ -78,7 +78,7 @@ describe('provider/bpmn - EscalationProps', function() {
         elementRegistry.get('EscalationIntermediateEvent_1')
       ];
 
-      escalationElements.forEach(async (element) => {
+      for (const element of escalationElements) {
 
         // when
         await act(() => {
@@ -87,9 +87,14 @@ describe('provider/bpmn - EscalationProps', function() {
 
         // then
         const select = domQuery('select[name=escalationRef]', container);
-        expect(select.value).to.eql(getEscalation(element).get('id'));
-      });
+        const escalation = getEscalation(element);
 
+        expect(select).to.exist;
+
+        if (escalation) {
+          expect(select.value).to.eql(escalation.get('id'));
+        }
+      }
     }));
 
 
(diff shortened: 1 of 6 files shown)


test/spec/provider/bpmn/EscalationProps.spec.js
```diff
@@ -78,7 +78,7 @@ describe('provider/bpmn - EscalationProps', function() {
         elementRegistry.get('EscalationIntermediateEvent_1')
       ];
 
-      escalationElements.forEach(async (element) => {
+      for (const element of escalationElements) {
```
nikku, line 81: :rofl:

test/spec/provider/zeebe/BusinessRuleImplementationProps.spec.js
```diff
@@ -366,20 +356,3 @@ function getTaskHeaders(element) {
 
   return getExtensionElementsList(businessObject, 'zeebe:TaskHeaders')[ 0 ];
 }
-
-async function expectEdited(container, exists) {
-
-  await wait(50);
```
nikku, line 372: Why would the `waitFor` utility from testing-library not work here, and keeping the test?

barmac: I don't know but it failed even after rewrite. I can try again today.

barmac: So this is how it looks with `waitFor`, but it fails on MacOS in chromeheadless: https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1111/commits/9abdfdbe04510ce67f694df4675c3864c207ee4d
It works in `.only` though.

barmac: I checked how the test suite behaves with different test suites, and it is green when I run only `provider` tests, but starts to fail when everything is included. Perhaps it's a slow clean up that jams on a local machine?

PR #1114, opened 2025-03-04
feat: add completionCondition + cancelRemainingInstances properties to ad-hoc subprocess

maff:
### Proposed Changes

Adds support to configure `completionCondition` (expression) and `cancelRemainingInstances` (boolean) for ad-hoc subprocesses.

**BPMN implementation**
image

**Zeebe specific implementation**
image

Related to https://github.com/camunda/camunda-modeler/issues/4850.

 

### Checklist

### DIFF
--- a/src/provider/bpmn/properties/AdHocCompletionProps.js
+++ b/src/provider/bpmn/properties/AdHocCompletionProps.js
@@ -6,9 +6,8 @@ import {
   TextFieldEntry,
 } from '@bpmn-io/properties-panel';
 
-import { createElement } from '../../../utils/ElementUtil';
-
 import { useService } from '../../../hooks';
+import { createOrUpdateFormalExpression } from '../../../utils/FormalExpressionUtil';
 
 /**
  * @typedef { import('@bpmn-io/properties-panel').EntryDefinition } Entry
@@ -52,9 +51,13 @@ function CompletionCondition(props) {
   };
 
   const setValue = (value) => {
-    return commandStack.execute(
-      'element.updateModdleProperties',
-      updateFormalExpression(element, 'completionCondition', value, bpmnFactory)
+    return createOrUpdateFormalExpression(
+      element,
+      getBusinessObject(element),
+      'completionCondition',
+      value,
+      bpmnFactory,
+      commandStack
     );
   };
 
@@ -97,60 +100,4 @@ function CancelRemainingInstances(props) {
     getValue,
     setValue,
   });
-}
-
-// helper ////////////////////////////
-
-// formal expression /////////////////
-
-/**
- * updateFormalExpression - updates a specific formal expression
- *
- * @param {djs.model.Base} element
- * @param {string} propertyName
- * @param {string} newValue
- * @param {BpmnFactory} bpmnFactory
- */
-function updateFormalExpression(element, propertyName, newValue, bpmnFactory) {
-  const businessObject = getBusinessObject(element);
-  const expressionProps = {};
-
-  if (!newValue) {
-
-    // remove formal expression
-    expressionProps[propertyName] = undefined;
-
-    return {
-      element,
-      moddleElement: businessObject,
-      properties: expressionProps,
-    };
-  }
-
-  const existingExpression = businessObject.get(propertyName);
-  if (!existingExpression) {
-
-    // add formal expression
-    expressionProps[propertyName] = createElement(
-      'bpmn:FormalExpression',
-      { body: newValue },
-      businessObject,
-      bpmnFactory
-    );
-
-    return {
-      element,
-      moddleElement: businessObject,
-      properties: expressionProps,
-    };
-  }
-
-  // edit existing formal expression
-  return {
-    element,
-    moddleElement: existingExpression,
-    properties: {
-      body: newValue,
-    },
-  };
 }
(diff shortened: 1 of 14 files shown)


src/provider/bpmn/properties/AdHocCompletionProps.js
```diff
@@ -0,0 +1,159 @@
...
+import { FeelEntryWithVariableContext } from '../../../entries/FeelEntryWithContext';
+
+/**
+ * @typedef { import('@bpmn-io/properties-panel').EntryDefinition } Entry
+ */
+
+/**
+ * @returns {Array<Entry>} entries
+ */
+export function AdHocCompletionProps(props) {
```
maff, line 22: I was first inclined to call this `CompletionProps`, but as these properties are only defined for ad hoc subprocesses I decided to go with this naming. Happy to adapt to a more generic naming to follow the pattern of the other props if needed.

src/provider/bpmn/properties/AdHocCompletionProps.js
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * updateFormalExpression - updates a specific formal expression
+ *
+ * @param {djs.model.Base} element
+ * @param {string} propertyName
+ * @param {string} newValue
+ * @param {BpmnFactory} bpmnFactory
+ */
+function updateFormalExpression(element, propertyName, newValue, bpmnFactory) {
```
maff, line 117: There is a certain amount of duplication across other properties which handle expressions - this one was mostly adapted from the `MultiInstanceProps`. Do we have a preference on duplication vs. extracting into reusable utils/helpers?

src/provider/bpmn/properties/AdHocCompletionProps.js
barmac: I think it's OK to have a file dedicated for ad-hoc subprocess. There's another place where we set the same property, but it's in a different context, so I'd rather have some duplication with adjustments possible than non-adjustable, shared entry.

image

src/provider/bpmn/properties/AdHocCompletionProps.js
barmac: Given that this is the third place where we can update formal expression (multiinstance, sequence flow condition, ad-hoc subprocess), having a util in https://github.com/bpmn-io/bpmn-js-properties-panel/tree/main/src/utils is a good idea.

maff: Refactored into a utility in 68d86a86dd896e8abe80aa03ba9caa38f7a0d3af and adapted the 2 new AdHoc props implementations to use that one. I would create a follow-up PR to adapt the other code parts as well to not expand the scope of this PR.

nikku: > Refactored into a utility in https://github.com/bpmn-io/bpmn-js-properties-panel/commit/68d86a86dd896e8abe80aa03ba9caa38f7a0d3af and adapted the 2 new AdHoc props implementations to use that one. I would create a follow-up PR to adapt the other code parts as well to not expand the scope of this PR.

As long as you clearly separate refactoring and your improvement in separate commits we're fine. We don't have limits in terms of size of a PR, what we ask for is that the commits are semantically structured. Your changes will also not be squashed, so whatever work you structure will be "followable" later on in the commit history.

maff: Fully agree on the semantically structured commits. I'm open to both - the follow-up PR would make it easier to follow the refactoring-only change on unrelated components with a dedicated PR description, but on the other hand changing other components here makes it easier to verify that the abstraction of the helper works properly for multiple use cases 😉 

I now updated the `MultiInstanceProps` to use the new helper in https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1114/commits/533f220dfed13a286d2b9666b94a341b25baea9a but kept the sequence flow as it was as its logic seems a bit different than the other 2.

barmac:
Is it the logical order in which I configure an ad-hoc subprocess? Initially, I thought that I'd first define which activities will be activated when element is reached, and only then I'd define the completion condition. However, the condition can be independent from the actions that the ad-hoc subprocess executor can take. I think the order is correct, but still want to flag it for discussion.
image

maff:
> Is it the logical order in which I configure an ad-hoc subprocess? Initially, I thought that I'd first define which activities will be activated when element is reached, and only then I'd define the completion condition. However, the condition can be independent from the actions that the ad-hoc subprocess executor can take. I think the order is correct, but still want to flag it for discussion.

💡 That's a very good point. I think they can independent, but it may be more logical to move the completion condition below the active elements when thinking at the flow of the process (first activate, then complete). I'll try to update it accordingly.

maff:
Updated to reorder the groups in https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1114/commits/1ab31141ed71c57f66f01fbd1af609c18edc512a:

image

barmac:
Thanks for the contribution. I think this is ready to merge. I've only pushed some formatting improvements (i.a. convention of 2 empty lines between tests): https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1114/commits/cdbadf07f269f64e7af14f22d3d8dbd28955d338

I will squash some of the commits, and merge it now.

barmac:
Released as [v5.34.0](https://github.com/bpmn-io/bpmn-js-properties-panel/blob/main/CHANGELOG.md#5340)

PR #1120, opened 2025-04-07
WIP Set parser dialect

Buckwich:
### Checklist

* [ ] **Brief textual description** of the changes present


src/provider/zeebe/properties/AdHocCompletionProps.js
```diff
@@ -60,6 +60,7 @@ function CompletionCondition(props) {
     id: 'completionCondition',
     label: translate('Completion condition'),
     feel: 'required',
+    feelParserDialect: 'camunda',
```
nikku, line 63: Is there any place where the `feelParserDialect` is not `camunda`? The reason I'm asking is that I think we should _inject_ the parser dialect, i.e. through a language context.

```
<FeelContext parserDialect="camunda" builtIns={ ... }>
  <AdHocCompletionProps ... />
</FeelContext>
```

barmac: This solution puts a burden on any future implementations to include `feelParserDialect` property. In the case of ZeebePropertiesProvider, it should always take `camunda`. So we should not require any new entries to have that property explicitly.

Buckwich: i've adjusted it to be configurable in one place, which can also be overwritten. but as we want to use the camunda dialect in all places in this project i set it as a default.

PR #1140, opened 2025-07-25
deps: update `@bpmn-io/properties-panel`

barmac:
### Proposed Changes

* `FEAT`: update `camunda` built-ins
* `FIX`: show tooltip on number fields
* `DEPS`: update to `@bpmn-io/properties-panel@3.31.0`

image

@Buckwich The documentation on the function sounds odd "in addition to previous overload" given that the ordering is reversed. Any idea what's going on? This is not a blocker.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to [bpmn-js-properties-panel](https://github.com/bpmn-io/bpm
 
 ___Note:__ Yet to be released changes appear here._
 
+## 5.39.0
+
+* `FEAT`: update `camunda` built-ins
+* `FIX`: show tooltip on number fields
+* `DEPS`: update to `@bpmn-io/properties-panel@3.31.0`
+
 ## 5.38.1
 
 * `FIX`: report Camunda user task as default implementation ([#1135](https://github.com/bpmn-io/bpmn-js-properties-panel/issues/1135))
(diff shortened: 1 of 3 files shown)


Buckwich:
It seems the completion popup doesn't care about the order we enter. The builtins contain them in the reverse order (ie. in the order from the docs). I also tested to manually rearrange them but codemirror rearranged them again.

We could think about utilising the boost functionality to give a better order or just don't rely on an order in the docs because we can never gurantee that all suggestions are shown (need to make other teams aware of that). Here's an example were the previous overload is not even shown: 
image

Buckwich:
@jarekdanielak do you see other issues than the fromAi function? that one was added recently to the docs without our constraints (no guarenteed order) in mind

Buckwich:
> I think a correct resolution would be to just keep descriptions focussed, and remove cross-references. This is however out of scope for this PR.

For reference: The source would need to be adjusted [here](https://github.com/camunda/camunda-docs/blob/main/docs/components/modeler/feel/builtin-functions/feel-built-in-functions-miscellaneous.md)

we need to make other teams aware how to write the docs

jarekdanielak (approved):
@Buckwich if the above issues are not blockers, I think we're good to go.

barmac:
> we need to make other teams aware how to write the docs

This or we have dedicated descriptions. I think it looks OK on the documentation page: https://docs.camunda.io/docs/next/components/modeler/feel/builtin-functions/feel-built-in-functions-miscellaneous/#fromaivalue
It's just that the description is not written with tooltip in mind.

PR #1215, opened 2026-04-22
feat: add multi-instance specific event types for execution listeners

AlekseyManetov:
Related to https://github.com/camunda/camunda-modeler/issues/5869

### Proposed Changes

Add multi-instance specific event types for execution listeners:
 - **Before All**: New type; applied to the XML as `beforeAll`.
 - **Before Each**: Renamed from 'Start'; applied to the XML as `start` (as before).
 - **After Each**: Renamed from 'End'; applied to the XML as `end` (as before).
 
The cleanup of the 'Before All' event when multi-instance is removed is done by this pr - https://github.com/camunda/camunda-bpmn-js-behaviors/pull/126 which is also incorporated here.
 
https://github.com/user-attachments/assets/f113e195-85f8-437c-8da7-dcb5e09bd786

### Checklist

Ensure you provide everything we need to review your contribution:

* [x] Contribution __meets our [definition of done](https://github.com/bpmn-io/.github/blob/main/resources/DEFINITION_OF_DONE.md)__
* [x] Pull request __establishes context__
  * [x] __Link to related issue(s)__, i.e. `Closes {LINK_TO_ISSUE}` or `Related to {LINK_TO_ISSUE}`
  * [x] __Brief textual description__ of the changes
  * [x] __Screenshots or short videos__ showing UI/UX changes
  * [ ] __Steps to try out__, i.e. [using the `@bpmn-io/sr` tool](https://github.com/bpmn-io/sr)

### DIFF
--- a/src/provider/zeebe/properties/ExecutionListener.js
+++ b/src/provider/zeebe/properties/ExecutionListener.js
@@ -1,6 +1,7 @@
 import { SelectEntry } from '@bpmn-io/properties-panel';
 
 import {
+  getBusinessObject,
   is,
   isAny
 } from 'bpmn-js/lib/util/ModelUtil';
@@ -22,6 +23,18 @@ export const EVENT_TO_LABEL = {
   'start': 'Start',
   'end': 'End'
 };
+
+// Specific event label for Multi instance elements: `beforeAll` runs once before MI init; `start` / `end` run per iteration.
+export const MI_EVENT_TO_LABEL = {
+  'beforeAll': 'Before all',
+  'start': 'Before each',
+  'end': 'After each'
+};
+
+export function getEventLabel(element, eventType) {
+  const labels = isMultiInstance(element) ? MI_EVENT_TO_LABEL : EVENT_TO_LABEL;
+  return labels[eventType];
+}
 
 export function ExecutionListenerEntries(props) {
 
@@ -78,7 +91,7 @@ function EventType(props) {
   const getOptions = () => {
     return eventTypes.map(eventType => ({
       value: eventType,
-      label: translate(EVENT_TO_LABEL[eventType])
+      label: translate(getEventLabel(element, eventType))
     }));
   };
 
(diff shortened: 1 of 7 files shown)


barmac:
> If a multi-instance element with a beforeAll event type is converted to a non-multi-instance element, we keep Before all in the event type dropdown as a disabled option to avoid showing an empty select value, and surface an inline validation error ("Please select a valid event type.") so the user notices the issue directly in the properties panel. Picking any valid event type clears the error. A corresponding lint error will also be added in bpmnlint-plugin-camunda-compat.

Isn't it better to clean up the invalid listeners ([do the hard work to make it simple](https://github.com/bpmn-io/design-principles#do-the-hard-work-to-make-it-simple))? This is what we do via http://github.com/camunda/camunda-bpmn-js-behaviors.

AlekseyManetov:
> Isn't it better to clean up the invalid listeners ([do the hard work to make it simple](https://github.com/bpmn-io/design-principles#do-the-hard-work-to-make-it-simple))?

Agree, that would be better. I initially chose the current solution to follow ["No surprises"](https://github.com/bpmn-io/design-principles#no-surprises), but [do the hard work to make it simple](https://github.com/bpmn-io/design-principles#do-the-hard-work-to-make-it-simple) is definitely a better fit here, especially if we have an established approach.

src/provider/zeebe/properties/ExecutionListener.js
```diff
@@ -115,5 +128,14 @@ export function getEventTypes(element) {
...
+  if (isMultiInstance(element)) {
+    return [ 'beforeAll', 'start', 'end' ];
+  }
+
   return [ 'start', 'end' ];
 }
+
+function isMultiInstance(element) {
+  const loopCharacteristics = getBusinessObject(element).get('loopCharacteristics');
+  return !!loopCharacteristics && is(loopCharacteristics, 'bpmn:MultiInstanceLoopCharacteristics');
```
barmac, line 140: Defensive approach, I like it.

barmac (approved):

barmac:
Works great

PR #1233, opened 2026-06-19
feat: add job priority properties for zeebe:jobPriorityDefinition

venetrius:
Adds a "Job priority" group with a single FEEL-optional `priority` field, shown on process, service task, send task, and (when implemented as a job worker) business rule task and script task.

Part of camunda/camunda-modeler#5889.

> Depends on the `zeebe:JobPriorityDefinition` moddle type (camunda/zeebe-bpmn-moddle) — needs the released moddle to build cleanly; tested here against a local overlay.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/provider/zeebe/ZeebePropertiesProvider.js
+++ b/src/provider/zeebe/ZeebePropertiesProvider.js
@@ -18,6 +18,7 @@ import {
   HeaderProps,
   InputPropagationProps,
   InputProps,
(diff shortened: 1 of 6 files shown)


barmac:
Done via https://github.com/bpmn-io/bpmn-js-properties-panel/pull/1235

